PR #49, opened 2016-03-17
tweaks to example book

jjallaire:
- split_by: chapter (that's the default and it works better with chapter preview)
- remove packages.bib from bibliography (it's .gitignored so results in an error)

### DIFF
--- a/inst/examples/_output.yml
+++ b/inst/examples/_output.yml
@@ -1,6 +1,6 @@
 bookdown::gitbook:
   css: css/style.css
-  split_by: section
+  split_by: chapter
   config:
     toc:
       collapse: subsection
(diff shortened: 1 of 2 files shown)


yihui:
I do not actually use `preview_chapter()` myself (via the addin Preview Book). I use `serve_book()` instead, so `split_by` is not an issue (I can split by anything).

For `packages.bib`, it is generated automatically from 06-references.Rmd, and you have to render the whole book at least once to generate this file (e.g. by compiling index.Rmd).

jjallaire:
Okay, I'm just thinking that if people clone the example as a starting
point then use RStudio Knit to preview they'll run into
difficulty/annoyance with split_by: section.

PR #53, opened 2016-03-18
Improvements to publish_book

jjallaire:
- Support for automatic updating of existing deployments with no arguments to publish_book (i.e. inherit the name, account, and server of the previous deployment so long as there is only one)

- Provide option to render book before publishing. This defaults to FALSE but can default to TRUE via the setting of a global bookdown option (this is for users that want to always render before publishing and don't want to have to remember to pass render = TRUE)

### DIFF
--- a/R/publish.R
+++ b/R/publish.R
@@ -8,22 +8,49 @@
 #' @param account Account name to publish to. Will default to any previously published
 #'  to account or any single account already associated with \code{server}.
 #' @param server Server to publish to (by default beta.rstudioconnect.com
-#' but any RStudio Connect server can be published to).
+#'  but any RStudio Connect server can be published to).
+#' @param render \code{TRUE} to render all formats prior to publishing (defaults to
+#'  \code{FALSE}, however this can be modified via the \code{bookdown.render_on_publish}
+#'  option). Note that this requires the use of a Makefile to provide the
+#'  implementaiton of rendering all formats.
 #'
 #' @export
 publish_book <- function(
-  name = NULL, account = NULL, server = 'beta.rstudioconnect.com'
+  name = NULL, account = NULL, server = NULL,
+  render = getOption("bookdown.render_on_publish", FALSE)
 ) {
 
   on.exit(opts$restore(), add = TRUE)
+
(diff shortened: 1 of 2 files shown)


R/publish.R
```diff
@@ -8,23 +8,50 @@
...
+  # render_book because we need some mechanism that will have a-priori
+  # knowledge of all book formats)
+  if (isTRUE(render)) {
+    if (file.exists("Makefile")) {
+      result <- system2("make")
+      if (result != 0)
+        stop("Error ", result, " attempting to render book")
+    }
+    else
+      stop("Rendering before publishing requires a Makefile")
```
yihui: Perhaps just call `render_book('index.Rmd')` here? Requiring Makefile may be an issue for Windows users.

jjallaire: Will render_book('index.Rmd') render all of the formats?

yihui: Oh right, that is an issue. In theory `render_book('index.Rmd', 'all')` should just work like `render()`, but I need to think more about it and try it.

jjallaire: Should we leave the Makefile in for now until we get
render_book('index.Rmd', 'all') working or should I remove the rendering
bit?

yihui: I think we can leave it there. If there is a Makefile, chances are the author knows what he/she is doing, and it has been documented anyway that `make` is expected to generate all formats.

jjallaire: Okay, sounds good. Let me know if there's anything else I should tweak here
before merging.

yihui: I thought a while about this, and instead of `render_all(output_format = 'all')`, I think a better approach is to look for an R script, say, `_bookdown.R`, in which user can write arbitrary code to compile the book to any formats. This is because 1) there may be multiple types of HTML or PDF output, e.g. using both `html_book` and `gitbook`, or two PDFs with different templates, in which case the user needs to make sure these output files do not overwrite each other; 2) currently the MOBI output is not really created from an output format, but simply by converting EPUB; in other words, it cannot be generated by `render_book()`, but it is easy to create a MOBI e-book using `bookdown::kindlegen()`.

PR #58, opened 2016-03-21
Change default server for publish_book to bookdown.org

jjallaire:


yihui:
Oh it is already up and running now! Nice :+1:

jjallaire:
I'm not sure when they will get the sitemap.xml and redirection to your
custom app from the "/" deployed, but I know that both features are
implemented so it shouldn't be too long now!

PR #59, opened 2016-03-23
implement custom rmarkdown site generators

jjallaire:
NOTE: Not ready for merge (requires this PR: https://github.com/rstudio/rmarkdown/pull/653)


jjallaire:
@yihui This is now ready for review. Note that I now propagate the "--quiet" flag to the _render.R script. 

I'm wondering if the site generator should have a fallback to render_book('index.Rmd', 'all') if there is no _render.R or Makefile?

yihui:
Yes, I think a fallback makes sense, but I'm not sure if `all` is a good default. I'd just use the default `output_format` passed from `render_site()` (which is `NULL` by default).

`--quiet` does not seem to be necessary for `Rscript`, since `Rscript` does not print the startup message (unlike `R`).

I think I'd prefer exporting `bookdown_site` and not creating the other three `_site` functions, which means users have to specify `site: bookdown::bookdown_site` instead of `site: true`. That may be a little bit hassle to users, but we will have fewer functions to maintain (although they are identical), and we will not have to think about adding other `foo_site` functions to this package in the future. I do not have strong opinions on this, though.

jjallaire:
Changes in response to feedback here: https://github.com/rstudio/bookdown/pull/59/commits/df991801f4ceeccada74c626116dff5ace904166

- rmarkdown now looks for a package level _site function if it can't find a format level one so bookdown_site will work.

- render_book is now used as a fallback default for the generator (forwarding the output_format)

- We do need to forward quiet -- that's actually not used for Rscript but rather for the calls to render inside the script (before it was just render = FALSE which meant that no output was ever shown during book rendering).

yihui:
Okay, that makes sense! Thanks!

PR #67, opened 2016-03-30
implement rmarkdown site generator clean function

jjallaire:
There is a new `clean` function which site generators must implement. It returns a site-relative list of files that should be cleaned (the `clean_site` wrapper function is the one that actually does the removal when the `preview` argument to `clean_site` is FALSE.

I was able to re-use the existing `clean_book` function by just arranging for it to return the `out` vector for `clean = FALSE`.


yihui:
I prefer `invisible(out)` there, and I'll do it myself. Thanks!

jjallaire:
Okay sounds good!

Issue #42, opened 2016-03-10
Think about gitbook performance


hadley:
e.g. https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fr4ds.had.co.nz&tab=desktop

I think the main problem is that the gitbook `app.js` is 760 kb!! This seems ridiculously large.

yihui:
Yeah, I'm aware of that. It will be minified eventually: https://github.com/rstudio/bookdown/blob/master/tools/uglifyjs.R And I could just take the embedded jQuery out.

PR #78, opened 2016-04-06
Take jQuery out from app.js

yihui:
This reduces the size of app.js from 750k to 500k (#42). I have tested it and it does not seem to break anything. @jjallaire Is this okay?

I'll minify app.js eventually to further reduce its size.


yihui:
The git diff is not displayed here. Basically I deleted these ~9000 lines: https://github.com/rstudio/bookdown/blob/6907616af74dfb7794fe6952a8722e5f73f1c611/inst/resources/gitbook/js/app.js#L2-L9212

jjallaire:
Looks good to me! Merge when ready. Note that rmarkdown now exports it's jquery dependency function so you could use that: https://github.com/rstudio/rmarkdown/blob/master/R/html_dependencies.R#L13-L21

Note however that it's using jquery 1.11.3 which is the same as the 2.2.3 version you are embedding but retains the IE 6,7,8 compatibility. I guess your dependency is probably better b/c it's smaller and gitbook doesn't support lower versions of IE anyway.

PR #98, opened 2016-04-29
close #69: add a clean_envir argument to clean the environment before rendering the book

### LINKED ISSUE #69: New session for each format when render_book(output_format = 'all')?
so that each format is generated from a clean R session

yihui:
@jjallaire Does this sound like a good idea?


jjallaire:
Yeah, I like this (especially with the default of `!interactive()`. It gives me some pause that we would be silently removing objects from the environment in batch mode but I honestly can't think of a real scenario where this would be of consequence.

PR #104, opened 2016-05-05
Update README.md

Nowosad:
`install_github` added


yihui:
Thanks! But I think users still have to read (at least the first few pages of) the documentation (https://bookdown.org/yihui/bookdown/) to really get started with this package, so I don't want to repeat the installation instructions here (otherwise when this package is published on CRAN in the future, I will have to update multiple places).

Nowosad:
Understood.

PR #150, opened 2016-07-08
document and demo text-based links

jennybc:
Thanks to #6 I was able to sort myself re: text-based links. Maybe you want to add this to the docs? I'm having trouble building that book, so you should check my links. But I have all of these forms working in a book that I can build, so I think I have them right.


yihui:
Sure. Thanks!

Issue #117, opened 2016-05-18
Treat blocks as markdown


yihui:
See https://github.com/jgm/pandoc/issues/2453

Instead of `\begin{block} \end{block}`, I can probably use other tokens so Pandoc can treat the block content as markdown, and I can restore the tokens to LaTeX environments in the end, e.g. translate this

    ```{block, type='foo'}
    a **block**
    ```

to

    \BookdownBegin{foo}

    a **block**

    \BookdownEnd{foo}

and Pandoc will convert it to

```latex
\BookdownBegin{foo}

a \textbf{block}

\BookdownEnd{foo}
```

then I restore it to

```latex
\begin{foo}

a \textbf{block}

\end{foo}
```

So all Pandoc's Markdown features can still be used in custom blocks.

Related #73.

safferli:
This would be awesome -- I'm currently trying to build a custom block which should include R code to be parsed. Would this be possible this way?

yihui:
I think the answer will be yes, if you mean you want R code chunks in custom blocks.

PR #154, opened 2016-07-13
Need to configure build tools to "Website"

dtkaplan:
Otherwise the "Build Book" button doesn't appear.

### DIFF
--- a/inst/examples/index.Rmd
+++ b/inst/examples/index.Rmd
@@ -61,6 +61,7 @@ The easiest way for beginners to get started with writing a book with R Markdown
     devtools::install_github('rstudio/bookdown')
     ```
 
+1. Bring up the `Build` tab of RStudio IDE. Follow the `More` menu and click to `Configure Build Tools`. Set `Project build tools` to `Website`. 
 1. Open the R Markdown file `index.Rmd` and click the button `Build Book` on the `Build` tab of RStudio;
 
 Now you should see the index page of this book demo in the RStudio Viewer. You may add or change the R Markdown files, come back to `index.Rmd`, and hit the `Knit` button again to preview the book. If you prefer not to use RStudio, you may also compile the book through command line. See the next section for details.


yihui:
Thanks! Actually I just added the project type yesterday https://github.com/rstudio/bookdown-demo/commit/382308f5a39690bbc70328c4a49f7199c3063802

dtkaplan:
Yihui,

I'm not sure what is the best way to communicate with you regarding ideas for bookdown, so I'll try this way.

First, I'm very excited about this project. It will be (and already is) a great collaboration tool.

I like to write my printed books in the tufte_book format.  In the past, I've just embedded latex commands like `\newthought{Start the paragraph}`, but now I'm trying to adopt to the more general style for non-PDF formats. I've been writing 
   `r tufte::newthought("Start the paragraph")` blah blah blah
which works, but is somewhat verbose.  On the other hand, your notation for references, `\@ref(fig:mylabel)` seems very nice. Might it be possible to extend this to other functions, including not having to put the contents in quotes (or escape quotes in the contents).  Like this
    \@newthought(Start the paragraph) blah blah blah
I realize that this is not much of a savings, but it would provide a nice mechanism to identify at a glance all R commands relating to the book format, e.g. `\@marginnote(My note is ...)`, `\@include_graphics(Images/myfile.png)` and so on.

On another topic, I've put together a primitive but effective system for translating from latex markup to Rmd. (I'm bringing a lot of existing materials into bookdown, and those were written in a combination of latex and markdown.) I have an addin that let's me replace any or all text in a document with latex commands translated into equivalent markdown. I don't know if this might be of interest to anyone but myself, but it's really useful for me!

yihui:
@dtkaplan I don't mind where you write the feedback, and I appreciate your input very much. I have exactly the same feeling as you: I prefer the syntax `\@command()` over `` `r function()` ``. I'll think more about it (#117 is related).

Re: translation from LaTeX to Markdown, yes, I'd love to see it, and I imagine it will be useful to LaTeX users who want to switch to Markdown.

PR #195, opened 2016-09-08
Added simpler method for publishing book to gh-pages via /docs folder

rudeboybert:
Using the `/docs` folder approach to publishing to gh-pages requires less steps than via a gh-pages branch.


yihui:
Fantastic! Thanks!

PR #212, opened 2016-09-24
Add author affiliation

samuel-rosa:
Add author affiliation in the YAML metadata block using the `institute` field. This field is available in other templates, such as in the default beamer template. See, for example, http://stackoverflow.com/a/35888123/3365410.

### DIFF
--- a/inst/templates/gitbook.html
+++ b/inst/templates/gitbook.html
@@ -27,6 +27,10 @@ $for(author-meta)$
 <meta name="author" content="$author-meta$">
 $endfor$
 
+$if(institute-meta)$
+<meta name="institute" content="$institute-meta$">
+$endif$
+  
 $if(date-meta)$
 <meta name="date" content="$date-meta$">
 $endif$


benmarwick:
Yes, this would be a very nice addition. I see it's in the [papaja](https://github.com/crsh/papaja) package for creating APA manscripts using Rmd. It would be great to have more YAML fields like this in bookdown because bookdown does captions and cross-referencing so well!

yihui:
Thanks! There may be a potential issue (actually not specific to this institute feature) -- could you quickly check what happens if the `institute` field contains special HTML characters like `<>&` or uses Markdown syntax like `**bold**`? The meta value should not contain HTML tags and special characters should be escaped.

samuel-rosa:
ok

Alessandro Samuel Rosa
Pedologia Quantitativa
Universidade Federal Rural do Rio de Janeiro

Visite meu website e deixe um comentário! <http://samuel-rosa.github.io>

Quer agendar uma atividade? Veja meu calendário!
<https://calendar.google.com/calendar/embed?src=alessandrosamuelrosa%40gmail.com&ctz=America/Sao_Paulo>

samuel-rosa:
@yihui, should I have added the `institute` field to `default.html` as well since that is the template file used to render gitbooks?

yihui:
`default.html` is irrelevant to the gitbook format. It is just the default template for `html_book`.

samuel-rosa:
I have noticed that the data passed to `intitute` is not being included in the resulting `gitbook`. Does anyone knows if there is something else that should be defined in the template?

yihui:
You can ask in the Pandoc mailing list if `institute-meta` is supported at all for HTML output.

PR #261, opened 2016-12-08
replaced "see" with "refer to"

Emaasit:
used "refer to" in place of "see", which appears again in a couple of words in the same sentence


yihui:
👍

PR #280, opened 2016-12-23
initial work on bookdown skeleton

kevinushey:
This PR adds a function, `bookdown_skeleton()`, as well as a bit of metadata that allows RStudio to use this function as a project template function (ie, ensure it's displayed within the `New Project...` dialog for users).

For now, I've basically just copied the content of the `bookdown-demo` project, from https://github.com/rstudio/bookdown-demo. Let me know if we should do something a bit more minimal here.

cc: @jjallaire @yihui, let me know if you have some thoughts on what the skeleton should look like.


jjallaire:
@yihui You should know that it's possible to present options for project creation to the user (see https://rstudio.github.io/rstudio-extensions/rstudio_project_templates.html). Let us know if you think there are options which we should include. Also let us know if you think a more minimal template should be used for creating a new project (i.e. less chapters by default).

yihui:
I think we can leave out the Travis-related files and toc.css.

It seems the daily Mac build of RStudio has been failing since 1.1.21?

kevinushey:
We finally have a fresh daily build of RStudio available, so if you have some free time soon please give it a shot!

I'll make the tweaks re: the Travis-related files + `toc.css`.

kevinushey:
Do we need the `DESCRIPTION` file as well? I notice that it confuses RStudio a bit (the project is detected as an R package rather than a bookdown project)

yihui:
You can get rid of DESCRIPTION. That is for Travis only.

yihui:
Let me merge the blogdown PR first (since it looks simpler) and play with it for a bit to familiarize myself with the project template stuff. I will come back to this PR later.

kevinushey:
Thanks! Please let me know as well if you have any questions, or if the documentation is unclear, or if you think there's something we should consider adding.

yihui:
It is crystal clear! Just tweaked the blogdown project template a bit and I'm very very happy to see this since we have really turned Hugo's 12 steps to build a website to 1 single step:

image

Perhaps the text input can be made a little wider?

R/skeleton.R
```diff
@@ -0,0 +1,21 @@
...
+
+  skeleton_files <- list.files(skeleton_resources,
+                               recursive = TRUE,
+                               include.dirs = FALSE)
+
+  source <- file.path(skeleton_resources, skeleton_files)
+  target <- file.path(path, skeleton_files)
+  file.copy(source, target)
+
+  TRUE
```
yihui: @kevinushey Is the returned value actually meaningful?

kevinushey: It currently is not, although it might make sense to allow for something here (e.g. a function to be called on startup for the new project, or something similar?)

jjallaire:
Agree that the text input should be wider. One general way to do this in a 2-column layout is for the caption for the text box to be *above* rather than to the left of the text box. A further refinement would be to allow the first column to span e.g. 65% of the widget if there is no second column. 

Finally, now that the new project dialog has absorbed all of this flexibility / features I think it would be completely fine to make it larger by default (there is some trickiness to this associated with fixed pixel sizes in the UI layout, but it should be straightforward enough to sort out).

kevinushey:
I can make a stab at expanding the dialog size today. When only one column is filled with outputs we indeed display the UI as though it were just a single column of widgets, so I think it will be sufficient to expand the size of the text widget here.

PR #284, opened 2017-01-02
Add R options to set buttons' text in html_book

walmes:
Use in R
```r
options(html_book_prev_text = "<some text>")
options(html_book_next_text = "<some text>")
```
to set the text on buttons "Previous' and 'Next'.

This PR enables user customize the text on buttons 'Previous' and 'Next' for another language and use ascii symbols (http://www.w3schools.com/charsets/ref_utf_dingbats.asp). It was done using R `options()` mechanism. It could be converted in arguments for `build_chapter()` or even yaml variables. I've tried to enable Font Awesome Icons (http://fontawesome.io/icons/) but without success.


yihui:
Thanks, but I hope this could be done in `_bookdown.yml`: https://bookdown.org/yihui/bookdown/internationalization.html

PR #364, opened 2017-03-13
numbering equations in word_document2

pzhaonet:
a trick to number and refer equations in word_document2 output.

### DIFF
--- a/inst/examples/02-components.Rmd
+++ b/inst/examples/02-components.Rmd
@@ -223,6 +223,20 @@ You can use the `split` environment inside `equation` so that all lines share th
 (\#eq:var-beta)
 \end{equation}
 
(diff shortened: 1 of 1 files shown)


inst/examples/02-components.Rmd
```diff
@@ -223,6 +223,20 @@ You can use the `split` environment inside `equation` so that all lines share th
 (\#eq:var-beta)
 \end{equation}
 
+If you would like a word_document2 output, you may find that the equation environment is not rendered, i.e. you probably cannot see the equation in the Word document. The easiest way to solve this problem is shown in the following example:
```
yihui: So you are essentially using the example list (http://pandoc.org/MANUAL.html#numbered-example-lists). It is a workaround, but I don't want to include it in the book since I don't think it is possible to adjust the alignment of the equation or the equation number. Users of bookdown are expected to read the Pandoc manual for at least once, and perhaps they can figure it out by themselves. I remember I saw this trick in the pandoc mailing list last year when I was searching for a solution to number equations in Word.

Thanks anyway for the contribution!

PR #371, opened 2017-03-20
Chemical formula typo

yixuan:
When using chemical formula to express ions, the digit precedes the `+` or `-` sign, for example, SO4<sup>2-</sup> and Fe<sup>2+</sup>. And when the digit is one, it is usually omitted, for example H<sup>+</sup> and Cl<sup>-</sup>.

(Oh no, do I really want to submit this as a PR... 😏)


yihui:
Let's replace it with `Fe^2+^` instead. I used `1` explicitly just to make sure the superscript more visually clear.

PR #374, opened 2017-03-20
Fix #373

### LINKED ISSUE #373: Extra white space around LaTeX environments
Pandoc automatically inserts blank lines on each end of a LaTeX environment, so
```
\begin{equation*}
\end{equation*}
```
results in
```

\begin{equation*}
\end{equation*}

```
in the .tex file that is compiled. This makes unwanted paragraph breaks in the final document. An  example of this can be seen in nicksolomon/thesisdown_example@a2dfda57b80f62011387d1b00899851188c2bf49

nicksolomon:
My fix for issue #373 that I just filed.


yihui:
This is a known issue of Pandoc: https://github.com/jgm/pandoc/issues/1196 I'd rather wait for it to be fixed upstream instead of hacking at it in bookdown. You are welcome to use the version you forked and modified before the issue is fixed in Pandoc (which appears unlikely to be fixed).

Besides, there are multiple issues with this PR. This appears to be your first PR (according to http://firstpr.me/#nicksolomon), and you may take this principle into consideration in the future when making contributions in the future:

> Do one thing and only one thing in a PR if possible.

For example, do not include the roxygen changes in this PR since they are distracting to reviewers (you need to update your roxygen2 package anyway). 

The most fundamental problem of this PR, though, is that the fix is too aggressive. You are essentially removing blank lines before/after **any** LaTeX environments. On one hand, the blank lines around some environments may not need to be removed; on the other, even if you target the fix to `equation*`, there is actually no way to tell if the blank lines are intentional or not, e.g. what if they are there in the R Markdown source? i.e. the author explicitly wanted the `equation*` environment to be a paragraph. I don't think it is possible to fix this issue from bookdown -- we have to rely on Pandoc to fix it.

Back to your original issue, you probably meant that the spacing was too large. I believe there must be LaTeX solutions to reduce the spacing, which should be much easier than hacking the LaTeX source generated by bookdown. That said, the spacing does not look too bad in my eyes: https://bookdown.org/yihui/bookdown/bookdown.pdf Or perhaps you mean the paragraph indentation?

nicksolomon:
Yes, this is my first pull request. Thanks a lot for such thoughtful feedback! I really appreciate it.

My trouble is coming from the paragraph indentation as the document I'm working on requires indented paragraphs with no extra space between them. So far, I've been putting `\noindent` after all my equations, but it's becoming a bit tedious. Hopefully a pandoc fix will come through sooner, rather than later.

Thanks again!

PR #402, opened 2017-05-30
Rename `window.require` in app.js (and app.min.js) to `gitbook.require`.

dataopt:
Rename `require` in all the plugin javascripts to `gitbook.require` and make
all the required changes to work with the renamed functions.

This renaming removes the naming conflict on the global require which is
needed by MathJax to activate the accessibility extensions.  Prior to this
renaming, MathJax will fail to load `[a11y]/mathjax-sre.js`, for example.


yihui:
Great. Thanks!

PR #404, opened 2017-06-02
Prefix section labels in cross-references 

aoles:
Prepend the value of `--id-prefix` passed to pandoc conversion in order to correctly resolve cross-references. Addresses https://github.com/rstudio/rmarkdown/issues/1063.


yihui:
Thanks!

PR #409, opened 2017-06-16
Fix for disappeared page navigation buttons

dataopt:
When using `split_by = section` with a hidden appendix section specified by `(APPENDIX)`, the section id of the appendix is prematurely deleted before the closing level 1 section div tag of the immediately preceding section is properly commented out when processing a level 2 section.  As a result, an extra closing `</div>` appears in the level 2 section, causing the prev-page and next-page button to not be rendered by the browser.

The fix consists of moving the div tag processing before the hidden appendix section is removed.


dataopt:
Has anyone tested this yet?

yihui:
I'm coming back to bookdown issues starting from today, and I'll look at this PR in the next few days. Thanks!

PR #428, opened 2017-07-26
Update 06-publishing.Rmd

atyre2:
Point out problem with default filename for pdf outputs with GitHub pages, and describe solution.


yihui:
Have you confirmed this with Github? It was only my guess: https://stackoverflow.com/a/45336202/559676 I can merge this PR if this behavior of Github Pages is by design.

yihui:
Wait a minute. I have already mentioned `.nojekyll` at https://bookdown.org/yihui/bookdown/github.html That might fix your original issue.

atyre2:
I have a .nojekyll file in the root directory of the repo. Does it need to be in the docs directory?

yihui:
Yes. It has to be where your website resides.

atyre2:
Ahhhh, yes. That works. So, maybe the bit mentioning .nojekyll should be more specific about where it goes if you are going down the docs route instead of the gh-pages branch route. Also, I think changing the default filename is a bit easier b/c if you clean the project you loose the .nojekyll file in docs and you have to put it back. I know, there's probably a place to tweak the rendering process to automatically add a .nojekyll file but in the interests of simplicity ...

yihui:
Please feel free to revert the change in this PR and mention `.nojekyll` for the `docs/` approach instead.

PR #429, opened 2017-07-27
add .nojekyll when doing /docs approach

atyre2:
Here is a 2nd try at explaining how .nojekyll fits in with the two different GitHub pages approaches.


yihui:
Thanks!

PR #461, opened 2017-09-08
Clean up after biber

jar1karp:
With `citation_package: biblatex` in `_output.yml` `biber` is used to process bibliographies nicely, but it leaves a `.run.xml` files lying around. This commit takes care of cleaning it up afterwards.

### DIFF
--- a/R/latex.R
+++ b/R/latex.R
@@ -62,7 +62,7 @@ pdf_book = function(
     if (is.function(post)) x = post(x)
     writeUTF8(x, f)
     latexmk(f, config$pandoc$latex_engine)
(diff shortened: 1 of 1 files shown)


R/latex.R
```diff
@@ -63,6 +63,7 @@ pdf_book = function(
     writeUTF8(x, f)
     latexmk(f, config$pandoc$latex_engine)
     unlink(with_ext(output, 'bbl'))  # not sure why latexmk left a .bbl there
+    unlink(with_ext(output, 'run.xml')) # clean up after biber
```
yihui: Could you merge this to the previous line? The second argument of `with_ext()` can take a vector.

jar1karp: Sure, done.

PR #465, opened 2017-09-21
Allow alternative config files

stephlocke:
I'd like to be able to have different config files instead of just `_bookdown.yml` - this enables me to have many files like `topicA`, `topicB`, `topic?` and be able to knit these into different books by pointing at different config file. This will help me reduce duplication and improve my ability to programmatically build books.

---
Notes on dev:

- I believe I've been stylistically consistent
- I haven't added any tests for alternative files but did confirm that all existing tests passed
- I'm in the process of completing the contributor agreement

### DIFF
--- a/R/ebook.R
+++ b/R/ebook.R
@@ -249,16 +249,17 @@ calibre = function(input, output, options = '') {
 #' @param exec The path to the executable \command{kindlegen}, which can be
 #'   downloaded from
 #'   \url{http://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211}.
+#' @param config_file File name (inc. relative path if required) of the configuration file for the render. The default is \file{_bookdown.yml}.
 #' @return The path of the \file{.mobi} file if the conversion is successful.
 #' @export
-kindlegen = function(epub, exec = Sys.which('kindlegen')) {
+kindlegen = function(epub, exec = Sys.which('kindlegen'), config_file="_bookdown.yml") {
   if (exec == '') stop(
     'Cannot find the executable KindleGen. You may download it from ',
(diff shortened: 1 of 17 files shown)


stephlocke:
Example repo now added https://github.com/stephlocke/bookdown-multiconfig

yihui:
Thanks! I don't quite like repeating this `config_file` argument everywhere. This will be a huge burden for me to maintain in the future.

I think a simple helper function can solve your problem:

```r
library(bookdown)

render_with_config = function(config) {
  if (file.exists('_bookdown.yml')) {
    tmp = tempfile()
    file.rename('_bookdown.yml', tmp)
    on.exit(file.rename(tmp, '_bookdown.yml'), add = TRUE)
  }
  file.rename(config, '_bookdown.yml')
  tryCatch(render_book("index.Rmd"), finally = {
    file.rename('_bookdown.yml', config)
  })
}

for (config in list.files(pattern = "*.yml")) {
  render_with_config(config)
}
```

The basic idea is to rename your config file to `_bookdown.yml` temporarily, and restore it after the book has been rendered. This way, you don't need to hack deep into the codebase of bookdown.

stephlocke:
Sneaky! I like it :) 

I'll close this PR. Would you be happy for your additional function to enter the bookdown codebase or just have it for ad-hoc use? If the function's OK to go in, I can create a new PR for it to save you the time.

yihui:
Thanks! I can do it myself this time :)

yihui:
I have added a `config` argument to `render_book()`.

stephlocke:
Thanks so much for integrating the request and the far more elegant solution!

Issue #497, opened 2017-12-14
Chapter and section numbering with preview_chapter (and related functions)


rundel:
I've attached a sample bookdown folder where the Rmds have been separated by section instead of chapter. [preview_bug.zip](https://github.com/rstudio/bookdown/files/1557745/preview_bug.zip)

This book renders correctly when using 
```r
render_book("index.Rmd", output_format = html_book(split_by="chapter"))
```
but I get misnumbering / mangling of the non-previewed sections when I attempt to use `preview_chapter`, `render_book` with `preview=TRUE` and `serve_book`.

For example, running the following with provide book
```r
preview_chapter("01-chap-02.Rmd", output_format = html_chapters())
```
gives me something that looks like

**Welcome**
**1 Chapter 1**
**2 Placeholder**
**2.1 Part 1**
**2.2 Part 2**
This is chapter one section two.
**3 Placeholder**
**3.1 Part 3**
**4 Chapter 2**
**5 Placeholder**
**5.1 Part 1**
`preview_chapter` (and hence `render_book`) is partially examining at least the first line of the other Rmds to come up with some of the placeholder text but it is not currently paying attention to the section level of that content. At least for the preceding sections it seems like it gets it partially right but there is still an additional unnecessary placeholder.

yihui:
Thanks for the report!

1. The documentation says (https://bookdown.org/yihui/bookdown/usage.html):

> [...] Each R Markdown file must start immediately with the chapter title using the first-level heading.

2. Preview is for, preview, only. https://bookdown.org/yihui/bookdown/preview-a-chapter.html You should only expect a single Rmd to be correctly previewed, and should not be surprised if funny things happen in the output of other Rmds.

rundel:
I appreciate that the documentation states that each Rmd should correspond to a single chapter but that is not always a reasonable workflow. There is no reason that the workflow described shouldn't work and it has been fine thus far when compiling the complete book, it is just the current preview related bits that are broken.

With regard to point 2. the existing code for preview is broken, hence the bug report. I've just submitted a pull request (#498) which attempts to address this. I haven't tested it extensively but it addresses the issues described above as well as an additional issue with code chunk output being interpreted as sections.

yihui:
Excellent! I'll review your PR tomorrow.

PR #498, opened 2017-12-15
Fix preview section numbering issue

rundel:
This is an attempt to fix the section numbering issue as reported in #497 as well as an additional issue with code output being interpreted as a section (due to the `##`).

I have complete removed the use of `# Placeholder` when an h3 is not found since this doesn't seem necessary (but I may have missed a use case).

In order to fix the code issue I've added an addition grep to identify code blocks and then checking for the `##` occuring inside or outside a block by counting the number of preceeding ```.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -143,7 +143,9 @@ merge_chapters = function(files, to, before = NULL, after = NULL, orig = files) 
   preview = opts$get('preview'); input = opts$get('input_rmd')
   content = unlist(mapply(files, orig, SIMPLIFY = FALSE, FUN = function(f, o) {
     x = readUTF8(f)
-    x = if (preview && !(o %in% input)) create_placeholder(x) else {
+    x = if (preview && !(o %in% input)) {
+      create_placeholder(x)
+    } else {
       insert_code_chunk(x, before, after)
     }
     c(x, '', paste0('<!--chapter:end:', o, '-->'), '')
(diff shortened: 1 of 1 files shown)


R/utils.R
```diff
@@ -143,7 +143,9 @@ merge_chapters = function(files, to, before = NULL, after = NULL, orig = files)
   preview = opts$get('preview'); input = opts$get('input_rmd')
   content = unlist(mapply(files, orig, SIMPLIFY = FALSE, FUN = function(f, o) {
     x = readUTF8(f)
-    x = if (preview && !(o %in% input)) create_placeholder(x) else {
+    x = if (preview && !(o %in% input)) {
+      create_placeholder(x)
+    } else {
```
yihui: In the future, please refrain from mixing cosmetic changes with essential changes in the same pull request.

R/utils.R
```diff
@@ -162,10 +164,24 @@ create_placeholder = function(x) {
...
+  code_blocks = grep('^```', x)
+
+  # Checks for ^#{2,} potentially occuring in a code block
+  # for each match count the preceeding occurances of ^```
+  #  - odd counts = in code block
+  #  - even counts = section / subsection / etc.
+  in_block = vapply(poss_sections, function(x) as.logical(sum(x>code_blocks) %% 2), TRUE)
+
+  sections = poss_sections[!in_block]
+  h4 = x[sections]
```
yihui: I don't have a strong opinion on this. I'm not against your improvement, but I don't mind if certain comments in the R code are treated as section headings, either. As I said, it is for preview only, and rough edges should be fine.

R/utils.R
```diff
@@ -162,10 +164,24 @@ create_placeholder = function(x) {
...
+
   c(
-    '', placeholder(head(h1, 1)), placeholder(head(h2, 1)),
-    placeholder(h3[1], '# Placeholder'), '', h4
+    '',
+    placeholder(head(h1, 1)), placeholder(head(h2, 1)), placeholder(head(h3, 1)),
```
yihui: So this is the actual fix to the ToC issue #497. There is indeed an edge case, for which I used `# Placeholder`, but the case is fairly minor, so I'm okay with removing the placeholder title.

R/utils.R
yihui: @rundel FYI: With `devtools::install_github(c('yihui/xaringan', 'rstudio/bookdown'))`, comments in fenced code blocks will no longer be treated as section headers.

rundel: 👍

PR #503, opened 2017-12-25
Worked around mysterious problems caused by pandoc 2.0.

dataopt:
The code in `html.R` seems to make assumptions on the html generated by pandoc that do not seem to be valid when pandoc version 2.0 is used.   This is PR is a work around that seems to work.


yihui:
Sorry, but could you provide a reproducible example first? I cannot take pull requests that only "seem" to fix issues.

PR #508, opened 2017-12-30
Fixes multiline footnote word-spacing issue.

dataopt:
Before the change, multiline footnote is not handled properly.  Consider the following minimal example:
```
---
title: "A Book"
author: "Frida Gomam"
site: bookdown::bookdown_site
documentclass: book
output:
  bookdown::gitbook: default
  bookdown::pdf_book: default
---

# Test {-}

Footnote test.^[Multiline
footnote
should
be
allowed]
```
Before the change, the entire footnote appears as one single word:
screen shot 2017-12-30 at 5 20 31 am

With the change, the footnote appears properly:
screen shot 2017-12-30 at 5 21 16 am

### DIFF
--- a/R/html.R
+++ b/R/html.R
@@ -847,7 +847,7 @@ parse_footnotes = function(x) {
   j = min(j[j > i])
   n = length(x)
   r = '<li id="fn([0-9]+)"><p>.+?<a href="#fnref\\1"[^>]*?>.</a></p></li>'
-  s = paste(x[i:n], collapse = '')
+  s = paste(x[i:n], collapse = ' ')
   items = unlist(regmatches(s, gregexpr(r, s)))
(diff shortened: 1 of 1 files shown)


R/html.R
```diff
@@ -847,7 +847,7 @@ parse_footnotes = function(x) {
   j = min(j[j > i])
   n = length(x)
   r = '<li id="fn([0-9]+)"><p>.+?<a href="#fnref\\1"[^>]*?>.</a></p></li>'
-  s = paste(x[i:n], collapse = '')
+  s = paste(x[i:n], collapse = ' ')
```
yihui: I guess `\n` is better than a space here.

dataopt: Done.

PR #557, opened 2018-03-24
Improve contrast of blockquote for default and sepia themes in gitbook output

dataopt:
The current font color settings for gitbook output for blockquote give an incorrect color in the sepia theme and also fail to achieve accessibility standards for both the default and sepia themes .  Before the change, we have the following:
screen shot 2018-03-24 at 5 28 46 am
screen shot 2018-03-24 at 4 54 13 am

A color test on the default theme shows that it fails to be WCAG AA normal text compliant as shown in the following:
screen shot 2018-03-24 at 5 29 18 am

After the change, we have:
screen shot 2018-03-24 at 6 05 26 am
screen shot 2018-03-24 at 6 05 57 am
which corrects the color in the sepia theme and attains WCAG AAA 18+ compliance and WCAG AA normal text compliance.
screen shot 2018-03-24 at 6 16 41 am


dataopt:
I thought https://github.com/rstudio/bookdown/commit/87b0a747cd7e98dab90a3ba8a96fb4cd324e292b wouldn't get lumped into this PR.  In any case, it adds Conjecture to the list of theorem-like environments since it is quite common to have conjectures labelled in the same way as propositions, lemmas and theorems.  If you prefer this change to appear in a separate PR, I will do a rewind.  Otherwise, you could consider approving it in this PR.

yihui:
Could you also add a news item in NEWS.md for the new conjecture environment? Thanks!

dataopt:
How I should add this to NEWS.md?  Do I add this to MINOR CHANGES under VERSION 0.7?

yihui:
It can be `NEW FEATURES` under 0.8.

dataopt:
Added Conjecture to NEWS.  Also included the required CSS addition.

PR #561, opened 2018-04-03
Allow `rmd_subdir` to be either a bool or list of subdirectories.

### LINKED ISSUE #242: Set specific subdirectory for RMD input files
This is a "feature request" as per [this Stack Overflow comment](https://stackoverflow.com/questions/40578609/set-r-bookdown-input-directory).

I'd like to specify a specific subdirectory to contain my `Rmd` files. I can include the specific file names in `_bookdown.yml` but I'd prefer not having to update that file every time a new .Rmd file is added.

If I set `rmd_subdir: true` then *all* `Rmd` files from the current directory and all subdirectories are included. However, I'd love to be able to specify a single subdirectory (or an array of subdirectories) for example by adding something like

    rmd_subdir: ["input/", "otherinput/"]

to `_boodown.yml`.

ellisvalentiner:
**Description:**

If `rmd_subdir` is true then bookdown searches for files *recursively* from the project root directory. If `rmd_subdir` is a list of subdirectories then bookdown search for files *recursively* from each subdirectory.

I have submitted the individual contributor agreement and welcome feedback on this pull request.

Fixes #242

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -17,7 +17,7 @@ Authors@R: c(
     person("David", "Shuman", role = "ctb"),
     person("Dean", "Attali", role = "ctb"),
     person("Drew", "Tyre", role = "ctb"),
-    person("Ellis", "Valentiner", role = "ctb")
+    person("Ellis", "Valentiner", role = "ctb"),
     person("Frans", "van Dunne", role = "ctb"),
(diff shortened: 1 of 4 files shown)


yihui:
Could you also post an answer to SO: https://stackoverflow.com/q/40578609/559676? Thanks!

PR #563, opened 2018-04-11
Fixes #60

### LINKED ISSUE #60: default HTML output on iPad
Hi @yihui 

Thanks
Kyle

dataopt:
This fixes the toolbar not appearing on iPad. (#60)


dataopt:
Unfortunately, the change alters the behaviour of toolbar.  However, the issue lies in the CSS for book-header fixed.

PR #564, opened 2018-04-11
Fixes #60

### LINKED ISSUE #60: default HTML output on iPad

dataopt:


dataopt:
I tested my books on a second-generation iPad mini running iOS 9, Firefox 59 on Mac and Windows 8.1, Safari on Mac, and IE 11 on Windows 8.1.

yihui:
Unfortunately the fix is probably not this simple. The `fixed` class was added to the book body for reasons. I don't remember all reasons now, but at least anchors won't work correctly without the `fixed` class (as in your fix)., e.g., currently they work correctly:

- https://bookdown.org/yihui/bookdown-demo/intro.html#fig:nice-fig
- https://bookdown.org/yihui/bookdown-demo/intro.html#tab:nice-tab

After your fix, they won't point to the expected elements (figure/table) on the page. The relevant CSS should be:
https://github.com/rstudio/bookdown/blob/b1b37d3d4bd4299f6103065d5d83c07307d9468c/inst/resources/gitbook/css/plugin-bookdown.css#L43-L45

dataopt:
So body-inner is getting in the way?  Would making it transparent work?

yihui:
I suspect this is a rather tricky issue specific to iPads. The fixed toolbar works fine on my iPhone (Safari).

dataopt:
This is too mysterious.  So it's not a webkit bug.

dataopt:
I found the root cause.  It's in the template `gitbook.html`.

PR #566, opened 2018-04-12
Fixes #60

### LINKED ISSUE #60: default HTML output on iPad

dataopt:
The template has incorrect nesting of the div for book-header.

### DIFF
--- a/inst/templates/gitbook.html
+++ b/inst/templates/gitbook.html
@@ -133,14 +133,14 @@ $endif$
 <!--bookdown:toc2:end-->
       </nav>
     </div>
-
+    
+    <div class="book-header" role="navigation">
+       <h1>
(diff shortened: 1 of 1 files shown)


dataopt:
In any case, the existing code is broken and iPad's behaviour cannot be regarded as buggy.  It's just by sheer luck that the code works almost everywhere else.

yihui:
`.book .book-header.fixed` means `.book-header` is a child of `.book` (the syntax for direct children is `parent > child`), but not necessarily the direct child, so that is not a conflict.

I mean this PR may well be correct, but we will certainly need to make other changes accordingly. It won't be as simple as moving the `book-header` div out from the `book-body`.

PR #570, opened 2018-04-16
allow custom pandoc templates for gitbook and epub

ThierryO:
The output only changes when the user explicitly provides a `template`. The output isn't changed when no `template` is given, making the changes backward compatible.

### DIFF
--- a/R/ebook.R
+++ b/R/ebook.R
@@ -15,6 +15,7 @@
 #' @param epub_version Whether to use version 3 or 2 of EPUB.
 #' @param md_extensions A character string of Pandoc Markdown extensions.
 #' @param pandoc_args A vector of additional Pandoc arguments.
+#' @inheritParams html_chapters
 #' @note Figure/table numbers cannot be generated if sections are not numbered
 #'   (\code{number_sections = FALSE}).
 #' @export
@@ -22,9 +23,9 @@ epub_book = function(
   fig_width = 5, fig_height = 4, dev = 'png', fig_caption = TRUE,
   number_sections = TRUE, toc = FALSE, toc_depth = 3, stylesheet = NULL,
   cover_image = NULL, metadata = NULL, chapter_level = 1,
-  epub_version = c('epub3', 'epub'), md_extensions = NULL, pandoc_args = NULL
+  epub_version = c('epub3', 'epub'), md_extensions = NULL, pandoc_args = NULL,
+  template
 ) {
-
   epub_version = match.arg(epub_version)
   args = c(
     pandoc_args,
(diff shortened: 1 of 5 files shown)


R/ebook.R
```diff
@@ -15,16 +15,17 @@
...
 #' @note Figure/table numbers cannot be generated if sections are not numbered
 #'   (\code{number_sections = FALSE}).
 #' @export
 epub_book = function(
   fig_width = 5, fig_height = 4, dev = 'png', fig_caption = TRUE,
   number_sections = TRUE, toc = FALSE, toc_depth = 3, stylesheet = NULL,
   cover_image = NULL, metadata = NULL, chapter_level = 1,
-  epub_version = c('epub3', 'epub'), md_extensions = NULL, pandoc_args = NULL
+  epub_version = c('epub3', 'epub'), md_extensions = NULL, pandoc_args = NULL,
+  template
```
yihui: Let's use `template = 'default'` here, and check if `!identical(template, 'default')` below.

R/ebook.R
```diff
@@ -15,16 +15,17 @@
 #' @param epub_version Whether to use version 3 or 2 of EPUB.
 #' @param md_extensions A character string of Pandoc Markdown extensions.
 #' @param pandoc_args A vector of additional Pandoc arguments.
+#' @inheritParams html_chapters
```
yihui: I prefer explicitly documenting this argument, and let users know the default EPUB templates can be found in the repo https://github.com/jgm/pandoc-templates

R/gitbook.R
```diff
@@ -5,13 +5,13 @@
...
 #'   Arguments to be passed to \code{rmarkdown::\link{html_document}()}
-#'   (\code{...} not including \code{toc}, \code{theme}, and \code{template}).
+#'   (\code{...} not including \code{toc}, and \code{theme}).
 #' @param config A list of configuration options for the gitbook style, such as
 #'   the font/theme settings.
 #' @export
 gitbook = function(
   fig_caption = TRUE, number_sections = TRUE, self_contained = FALSE,
-  lib_dir = 'libs', pandoc_args = NULL, ...,
+  lib_dir = 'libs', pandoc_args = NULL, ..., template = bookdown_file('templates', 'gitbook.html'),
```
yihui: I think it is also worth documenting the `template` argument explicitly. In particular, we need to warn users that you have to fully understand the default template before using a custom template for gitbook. I don't know what your use case here. I feel it should be very rare that one needs a custom template for gitbook.

ThierryO: I want to use extra pandoc variables to place certain elements at a specific location which specific formatting at the title page to match our corporate identity style. The [template](https://github.com/inbo/INBOmd/tree/cd50301646272571482058265ed20bef3c6e5d4f/inst/rmarkdown/templates/inbo_rapport/skeleton) will be used by people less knowledgeable on bookdown.

I'll make the requested changes. I'll recommend to start from the [gitbook.html template](https://github.com/rstudio/bookdown/blob/master/inst/templates/gitbook.html) when creating a custom template.

PR #571, opened 2018-04-19
Document section headings without numbering

ms609:
Adding the comment "If you want the section heading to be unnumbered, add `{-#id}`."

It took me a fair bit of trial and error to figure this out, so I reckon it might deserve a line in the docs.


ms609:
Yes, I'd seen `{-}`, but wasn't sure how to combine it with a link.

I've found it useful to give an unnumbered section a distinct name, as then I can link to that section by hyperlink from elsewhere in the document with a friendly and stable URL -- e.g. I can access the unnumbered section "Supplementary table" with a link to `(#table)` in https://github.com/ms609/hyoliths/blob/master/index.Rmd; that link then takes me to the concisely labelled https://ms609.github.io/hyoliths/table.html

yihui:
I see. This won't work for printed versions, because the link is only accessible in electronic formats. In the context where you added this comment, readers may think the syntax `\@ref()` works. Actually even I thought so when I saw your PR, but it doesn't work. So if we want to mention this in the book, this paragraph may not be the appropriate place.

I've mentioned several times that if you are serious about Markdown, you need to read Pandoc's manual at least once: https://yihui.name/en/2017/11/thanks-tj-mahr/ Specifically, you wouldn't have to go through the trial and error if you had read this section in the manual: http://pandoc.org/MANUAL.html#header-identifiers

In conclusion, I'm open to adding this comment to the book, but we need to make it clear that `\@ref()` won't work (perhaps add a third bullet to https://bookdown.org/yihui/bookdown/cross-references.html?). BTW, I slightly prefer the syntax `{#ID .unnumbered}` (at least it should also be mentioned); `{-#ID}` may be a little obscure albeit compact.

Thanks!

ms609:
With this in mind, perhaps the best thing is to link to Pandoc's own documentation here, to avoid duplicating it and to provide advanced users with a direct link to all they might possibly need to know on this topic.  I've put a proposed wording in the latest commit.

yihui:
Thanks!

PR #577, opened 2018-05-09
Add recursive = TRUE to unlink functions in render.R

flyaflya:
The creation of temp files using the `tempfile()` function creates files with a path of ".\\\filename".  However, when using the `unlink `function with default of `recursive = FALSE`, these created temp files are not removed as expected.  When `recursive = TRUE` is added to the unlink function call, then it works.  Code which reproduces the problem on my machine is below:

```r
## mimic creation of tempFile name as is done in bookdown::render.R
tempFile = tempfile('render', '.', '.rds')
print(tempFile)
file.exists(tempFile)  ### returns FALSE, nothing written to file

## write something to file
writeLines("This is a temp file", con = tempFile)
file.exists(tempFile)  ### returns TRUE

## use unlink with default value of recursive = FALSE
unlink(tempFile)

## check if file exists
file.exists(tempFile)  ### returns TRUE, file still exists!

## use unlink with non-default value of recursive = TRUE
unlink(tempFile, recursive = TRUE)

## check if file exists
file.exists(tempFile)  ### returns FALSE, finally got rid of it

## I do not know why this option needs to be TRUE as it does not
## seem we are doing a recursive search down a directory tree.
## My only guess is it has something to to with the path specification
## as tempFile = ".\\render2b885fcas674b.rds" and I am not familiar
## with the ".\\" notation.
```

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -154,7 +154,7 @@ render_new_session = function(files, main, config, output_format, clean, envir, 
 
   # save a copy of render arguments in a temp file
   render_args = tempfile('render', '.', '.rds')
-  on.exit(unlink(render_args), add = TRUE)
+  on.exit(unlink(render_args, recursive = TRUE), add = TRUE)
   saveRDS(
     list(output_format = output_format, ..., clean = FALSE, envir = envir),
     render_args
@@ -186,7 +186,7 @@ render_new_session = function(files, main, config, output_format, clean, envir, 
       write_utf8(txt, f)
       Rscript_render(f, render_args, render_meta)
     }, finally = {
-      if (file.copy(f2, f, overwrite = TRUE)) unlink(f2)
+      if (file.copy(f2, f, overwrite = TRUE)) unlink(f2, recursive = TRUE)
     })
   }
   if (!all(dirname(files_md) == '.'))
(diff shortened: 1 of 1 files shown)


yihui:
So basically `unlink()` is unable to delete a file on your system, which is surprising to me. Does `file.remove()` work?

flyaflya:
I debugged a little more and it turns out having something to do with my use of OneDrive.  It turns out that if the file is created in a OneDrive synced working directory, `unlink(recursive = FALSE)` will NOT remove it.  When I create the file in a non-OneDrive folder, `unlink(recursive = FALSE)` will work.  File.remove, in addition to `unlink(recursive = TRUE)`, does seem to work in both situations.  Dropbox has no problems.  

PS:  Thanks for the path notation answer.  My true confusion with the path is the double backslash ".\\render...".  I would have just used a single backslash when defining the file path.

yihui:
In R (and many other languages), when you see double backslashes in a character string, it means a single literal backslash (yes, this is confusing forever: https://xkcd.com/1638/).

One more test I need you to help me with: does `unlink()` work when path does not contain the leading `.\`? In your example above, change `tempFile` to

```r
tempFile = basename(tempfile('render', '.', '.rds'))
```

flyaflya:
The behavior is unchanged with `tempFile = basename(tempfile('render', '.', '.rds'))`.  The issue with the OneDrive synced directory remains.

Thanks for the cartoon ... makes me feel better about my confusion :-)

PR #599, opened 2018-07-11
Solve error without .Rprofile

bubifengyun:
Error in contrib.url(repos, type) :
trying to use CRAN without setting a mirror
Calls: ... lapply -> FUN -> install.packages -> grep -> contrib.url

if without .Rprofile, the original code can pop the same error.

### DIFF
--- a/inst/examples/index.Rmd
+++ b/inst/examples/index.Rmd
@@ -26,9 +26,10 @@ options(
   width = 55, digits = 4, warnPartialMatchAttr = FALSE, warnPartialMatchDollar = FALSE
 )
 
-local({r <- getOption("repos")
-  r["CRAN"] <- "https://cran.rstudio.com/" 
-  options(repos=r)
+local({
+  r = getOption('repos')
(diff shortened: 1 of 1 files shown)


bubifengyun:
oh, it should add 

if r["cran"] == null then set r["cran"] = default website， I will try again

yihui:
Actually I don't understand how this error could occur. I set the `repos` option conditional on `!length(getOption('repos'))`, which means I'll set `repos` only if you didn't set it. If you have set it, you have to make sure you have set it correctly. Please show your `getOption('repos')`.

bubifengyun:
My OS is centos 7.4 min install x64 as server. result is 

```
> getOption('repos')
    CRAN 
"@CRAN@"
```

I think it is not so easy.

bubifengyun:
I had tested, it seems not works well

```
> r = getOption('repos')
> r
    CRAN 
"@CRAN@" 
> identical(r['CRAN'], '@CRAN@')
[1] FALSE

```

maybe,

```
if (!length(r) || all(r['CRAN'] == '@CRAN@')) r['CRAN'] = 'https://cran.rstudio.com' 
```

PR #601, opened 2018-07-13
Allow both `rmd_files` and `rmd_subdir` config parameters

### LINKED ISSUE #600: LaTeX template not respected when YAML option rmd_subdir is used with non-boolean value
Thanks for this package. I tried to use the rmd_subdir option introduced in #561. However, I encounter the following problem: 
When I specify a latex template (I took the default pandoc template) for the book, I lose it as soon as I use the option `rmd_subdir: ["rmd"]`. I forked and modified the [bookdown demo](https://github.com/rstudio/bookdown-demo) to create a reprex, which you can find [here](https://github.com/lorenzwalthert/bookdown-demo). Just see the last few commits. See https://github.com/lorenzwalthert/bookdown-demo/commit/d9bcee01fe7a78c505f90d618dda8a832858d311#diff-b5352b26567fa5f42ad7241975ba203bR6 for the commit where I added the option. If you remove the line and re-build the book, you will see that the template again takes effect. Otherwise the option seems to work fine. I don't understand what's going on. Any help appreciated.

ellisvalentiner:
Fixes:

- :bug: Allows `rmd_subdir` and `rmd_files` config parameters in `_bookdown.yml` to be used simultaneously. Previously `rmd_files` would overwrite the `files` list containing the files in the `rmd_subdir` paths.

Notes:

- `index.Rmd` is likely in the project root directory and not in a subdirectory in `rmd_subdir` so it is probably a good idea to specify `rmd_files: ["index.Rmd"]` if also using `rmd_subdir`.

I have submitted the individual contributor agreement and welcome feedback on this pull request.

Fixes #600


lorenzwalthert:
Works for me to resolve #600

R/utils.R
```diff
@@ -64,7 +64,7 @@ source_files = function(format = NULL, config = load_config(), all = FALSE) {
     recursive = subdir_yes, full.names = subdir_yes
   )
   if (length(config[['rmd_files']]) > 0) {
-    files = config[['rmd_files']]
+    files = c(config[['rmd_files']], files)
```
yihui: The concatenation needs to be done after the next `if` (i.e. after the list is resolved to a character vector).

R/utils.R
```diff
@@ -75,6 +75,7 @@ source_files = function(format = NULL, config = load_config(), all = FALSE) {
     if (!is.na(index)) files = c(files[index], files[-index])
   }
   check_special_chars(files)
+  unique(normalize_path(files))
```
yihui: We need to keep the file paths relative and should not resolve them to absolute paths. But you are right that we should make sure the file paths are unique.

Gnossos:
I tried using the approach in the Notes at the start of this topic but ran into trouble. In my project's root directory I have another file, Notes.rmd, which I use to keep track of things I do to configure and maintain my workflow. In _bookdown.yml I have:
    rmd_files: ["index.Rmd"]
    rmd_subdir: ["chapters"]
And when I build the book, the build includes Notes.Rmd in the book. I had assumed this setup would only include files explicitly named in rmd_files: and search only in chapters. obviously this is wrong. Is there a way to exclude Notes.Rmd from being in the built book?

cderv:
Please can you open a new issue, referencing this one, to discuss your specific case ? 
It is better for us to keep track of new issue in separate GH ticket and not mixed discussions. Thank you. 

Also, if `Notes.Rmd` is a file you don't want, you can prefix with `_Notes.Rmd` and it will be ignored by default (no need of `rmd_files` in this case) See https://bookdown.org/yihui/bookdown/usage.html#usage
> By default, bookdown merges all Rmd files by the order of filenames, e.g., 01-intro.Rmd will appear before 02-literature.Rmd. Filenames that start with an underscore _ are skipped

PR #603, opened 2018-07-18
Add support to tintBook

lf-araujo:
Dear authors, I am new to git, so forgive me if I did it wrong. 

I noticed that with two local file changes, one can add support to eddelbuttel tintBook, so I made the changes. It works in my local tests with Manjaro linux. 

All the best

https://github.com/eddelbuettel/tint

PR #604, opened 2018-07-18
Add support to tintBook

lf-araujo:
Same as previous, added to NAMESPACE


yihui:
Please see #603. Thanks!

Issue #612, opened 2018-08-11
Change name of temporary files created during build process


jrnold:
Bookdown leaves behind temporary files with names like `bookdown4def523aea53`.  This isn't a problem directly, but the names are hard to match with globbing rules, like those used in `.gitignore`.  It's hard to do be much more specific than `/bookdown*`. But that pattern could result in false positives. Could these temp file names be made more distinctive?

```{r}
> devtools::session_info('bookdown')
Session info -----------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.5.1 (2018-07-02)
 system   x86_64, darwin15.6.0        
 ui       RStudio (1.1.456)           
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       America/Los_Angeles         
 date     2018-08-11                  

Packages ---------------------------------------------------------------------------------------------
 package   * version date       source                            
 base64enc   0.1-3   2015-07-28 CRAN (R 3.5.0)                    
 bookdown  * 0.7.17  2018-08-10 Github (rstudio/bookdown@02b0fd1) 
 digest      0.6.15  2018-01-28 CRAN (R 3.5.0)                    
 evaluate    0.11    2018-07-17 cran (@0.11)                      
 glue        1.3.0   2018-07-20 Github (tidyverse/glue@66de125)   
 graphics  * 3.5.1   2018-07-05 local                             
 grDevices * 3.5.1   2018-07-05 local                             
 highr       0.7     2018-06-09 CRAN (R 3.5.0)                    
 htmldeps    0.1.1   2018-08-10 Github (rstudio/htmldeps@c1023e0) 
 htmltools   0.3.6   2017-04-28 CRAN (R 3.5.0)                    
 jsonlite    1.5     2017-06-01 CRAN (R 3.5.0)                    
 knitr     * 1.20    2018-02-20 CRAN (R 3.5.0)                    
 magrittr  * 1.5     2014-11-22 CRAN (R 3.5.0)                    
 markdown    0.8     2017-04-20 CRAN (R 3.5.0)                    
 methods   * 3.5.1   2018-07-05 local                             
 mime        0.5     2016-07-07 CRAN (R 3.5.0)                    
 Rcpp        0.12.18 2018-07-23 cran (@0.12.18)                   
 rmarkdown   1.10.11 2018-08-10 Github (rstudio/rmarkdown@adb0279)
 stats     * 3.5.1   2018-07-05 local                             
[...]

PR #613, opened 2018-08-12
Change name of temporary files created.

jrnold:
Currently, during rending the code copies the backs up the R markdown file to a temporary file in the same directory with the a name like `bookdown11b6326822f9c`. This file should be cleaned up, but sometimes is not if the process is killed. If the file happens to not be cleaned up, it is hard to write a globbing expression for `.gitignore` and similar to match it that and wouldn't result in false positives; `bookdown*` is the best I can come up with. I propose

1. Use a `.bak` extension to make it easier to match with globbing.
2. Use the original filename as the base, since it is backing up a particular file. If the process is killed, then it is easier to restore the original file.

Alternatively, if the backup file never needs to be restored by the user, why not put it in the temporary directory?

Addresses #612

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -178,7 +178,7 @@ render_new_session = function(files, main, config, output_format, clean, envir, 
       next
     }
     # first backup the original Rmd to a tempfile
-    f2 = tempfile('bookdown', '.')
+    f2 = paste0(f, '.bak')
     file.copy(f, f2, overwrite = TRUE)
     # write add1/add2 to the original Rmd, compile it, and restore it
     tryCatch({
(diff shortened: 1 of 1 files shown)


jrnold:
That's fine with me; the new commit makes that change. Is there a reason the backup file has to be in the working directory? It's only used for backup and not execution.

yihui:
To make it easier to restore the file if anything goes wrong (e.g. R crashed). If it is in the `tempdir()`, I'm not sure it can always be preserved, and it is usually hard to find.

jrnold:
👍

PR #630, opened 2018-09-12
Update style.css to make customizeable

technocrat:
Broke the css line by inserting \n before each { to make it possible to see what the code does and to modify it. I'd prefer leaving it as is, but I couldn't find any way to override style.css as the default.


yihui:
Got you. Usually what I do is use Inspect Element in Chrome, find the CSS selector, and rewrite its rules in `custom.css`. And I don't need to look at gitbook's minimized CSS.

technocrat:
Ah! Keys to the kingdom. Profuse thanks.



On September 12, 2018 at 11:38:31 AM, Yihui Xie (notifications@github.com)
wrote:

Got you. Usually what I do is use Inspect Element in Chrome, find the CSS
selector, and rewrite its rules in custom.css. And I don't need to look at
gitbook's minimized CSS.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rstudio/bookdown/pull/630#issuecomment-420754269>, or mute
the thread

Issue #638, opened 2018-09-27
How to add an history button


maelle:
Thanks for a great package and sorry for asking here. I fear this is more a feature request than a question. :wink:

Is it already possible to add an history button for the gitbook format, that'd be like an edit button but pointing at the history of the source instead, and with a different icon? (maybe [fa-history](https://fontawesome.com/v4.7.0/icon/history)) 

If not, I'd be happy to try and make a PR for that.

PR #639, opened 2018-09-29
Add option for a `history` button.

noamross:
This PR adds the ability to add a "History" button, similar to an "Edit" button, as part of custom configuration.  It essentially duplicates the code associated with the edit button code so that you can get something like this:

screenshot 2018-09-29 16 38 14

I've included changes to the gitbook documentation but not in the examples.

This addresses #638 from @maelle. I note I previously signed an individual contributor agreement on 2017-07-28.


noamross:
I've collapsed those functions to `source_link()` and `source_link_setting()`, both with a 'type' option, squashed to a single commit and rebased to master.

R/gitbook.R
```diff
@@ -138,10 +138,16 @@ gitbook_page = function(
   }
 
   # you can set the edit setting in either _bookdown.yml or _output.yml
-  if (is.list(setting <- edit_setting(config$edit))) config$edit = setting
+  if (is.list(setting <- source_link_setting(config$edit, type="edit")))
```
yihui: nit: spaces around `=`

R/gitbook.R
```diff
@@ -138,10 +138,16 @@ gitbook_page = function(
...
+    config$edit = setting
   if (length(rmd_cur) && is.list(config$edit))
     config$edit$link = sprintf(config$edit$link, rmd_cur)
 
+  if (is.list(setting <- source_link_setting(config$history, type="history")))
+    config$history = setting
+  if (length(rmd_cur) && is.list(config$history))
+    config$history$link = sprintf(config$history$link, rmd_cur)
```
yihui: I think the above two blocks of code can also be collapsed into a for loop. They look too similar.

```r
for (type in c('edit', 'history')) {
  ...
}
```

R/html.R
```diff
@@ -430,21 +431,21 @@ button_link = function(target, text) {
   )
 }
 
-edit_link = function(target) {
+source_link <- function(target, type) {
```
yihui: `<-` -> `=`

noamross:
Sorry for the delay! Just addressed these changes and collapsed to a single commit.

Issue #641, opened 2018-10-04
[feature request] Allow users to choose between gitbook tables & bootstrap tables


haozhu233:
gitbook's table style is quite basic comparing with tables in bootstrap. I'm wondering if it's a good idea to give people option to switch between these two styles. Right now the entire gitbook css is loaded as a whole chunk. Even if I load bootstap table style after, since existing css settings cannot be reverted, some bootstrap features (like hovering effects) cannot be displayed correctly (See https://haozhu233.github.io/kableExtra/bookdown/use-bootstrap-tables-in-gitbooks-epub.html#gitbook). 

I'm thinking if it's possible to make some modifications to the gitbook css and separate out the table part. At the same time, put bootstrap 3's table part (See [this](https://getbootstrap.com/docs/3.3/customize/) or [that](https://github.com/haozhu233/kableExtra/blob/master/inst/bootstrapTable-3.3.7/bootstrapTable.min.css)) into the resource list and create some options to switch between these two in [gitbook_dependency](https://github.com/rstudio/bookdown/blob/b6b96b629c78e71f0ed47db9bfead389ba5fe09a/R/gitbook.R#L75-L90). Both gitbook and bootstrap are under MIT so I guess it's okay to mix them two together. 

```
bookdown::gitbook:
    bootstrap_table: true/false
```

PR #642, opened 2018-10-04
#641 allow users to choose between gitbook table and bootstrap table

haozhu233:
I tried to implemented what I described in #641. Please feel free to reject it or let me know what needs to be changed. I will add a NEWS item after approval.


haozhu233:
Sounds good, I will make some updates this afternoo.

haozhu233:
@yihui I just made the changes according to your comments. I feel like this is even better now as people can use whatever css they would like to use. Thanks for your suggestion!

yihui:
Well, it seems you have got more work to do: after this PR, there is an undesired change in the table style. See Table 2.3 in the bookdown book: https://bookdown.org/yihui/bookdown/tables.html

Before the PR:

image
After the PR:

image

I'd like to get rid of the vertical lines.

PR #647, opened 2018-11-01
Added label to search form for screen readers

rblades:
Appended HTML <label> to $searchForm for increased accessibility. The label will help screen readers that may not read the <input> placeholder attribute (in this case, "Type to search"). The procedure to add a <label> element alongside an <input> is suggested by Khan Academy's [tota11y accessibility tool](https://github.com/Khan/tota11y). Refer to [W3C on labels and web accessibility](https://www.w3.org/WAI/tutorials/forms/labels/).


rblades:
Awesome, thanks so much!

PR #672, opened 2019-01-19
Add support for Hungarian language @aref

### LINKED ISSUE #671: How to handle a specialized addition to bookdown?
[This question](https://stackoverflow.com/q/54154890/2554330) on StackOverflow asks for a very specialized addition to `bookdown` to support the Hungarian language:  rather than `\ref` in LaTeX, it needs `\aref` from the `babel-magyar` package (or something like that), because Hungarian has special grammatical rules for dealing with numbers.

I wrote a [tiny package](https://github.com/dmurdoch/magyarbookdown) that monkey-patches bookdown when loaded to support `@aref` as well as `@ref`.  It's a pretty fragile solution, because as soon as `bookdown` updates one of the patched functions, my package will also need an update.  I could submit a PR to incorporate the code into the main `bookdown`, but I suspect there aren't really that many Hungarian users, and I doubt if the `bookdown` authors want to support this.  And there are a lot of languages, so there are probably other specialized patches that others would want.

So what would be a good strategy for going forward with this?

dmurdoch:
As mentioned in Issue #671, the babel Hungarian support in LaTeX adds the macro `\aref` to be like `\ref` but with support for some special Hungarian grammatical constraints.

This PR takes the changes from my [magyarbookdown](https://github.com/dmurdoch/magyarbookdown) package and includes them in `bookdown`.   I'm not sure this is an appropriate addition, but I'll let you decide.

### DIFF
--- a/R/ebook.R
+++ b/R/ebook.R
@@ -126,8 +126,8 @@ resolve_refs_md = function(content, ref_table, to_md = output_md()) {
   # replace (\#eq:label) with equation numbers
   content = add_eq_numbers(content, ids, ref_table, to_md)
 
-  # look for \@ref(label) and resolve to actual figure/table/section numbers
-  m = gregexpr('(?<!`)\\\\@ref\\(([-:[:alnum:]]+)\\)', content, perl = TRUE)
+  # look for \@ref(label) or \@aref(label) and resolve to actual figure/table/section numbers
+  m = gregexpr('(?<!`)\\\\@a?ref\\(([-:[:alnum:]]+)\\)', content, perl = TRUE)
   refs = regmatches(content, m)
   regmatches(content, m) = lapply(refs, ref_to_number, ref_table, TRUE)
   content
--- a/R/html.R
+++ b/R/html.R
@@ -453,7 +453,8 @@ source_link_setting = function(config, type) {
 #'
 #' Post-process the HTML content to resolve references of figures, tables, and
 #' sections, etc. The references are written in the form \code{\@ref(key)} in
-#' the R Markdown source document.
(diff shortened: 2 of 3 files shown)


yihui (commented):
With the recent PR #1120 (aiming to fix #1114), it's possible to customize figure labels via a function provided by users. I don't know anything about Hungarian, but I guess at least it should be possible to generate a label like `az 1. ábra` now for non-LaTeX output. If that's the case, we don't need to embed the label-generating logic for a particular language inside this package.

For LaTeX output, perhaps we just need a config option for users to specify which LaTeX command they'd like to convert `\@ref()` to (the default would still be `\ref{}`). Again, I know nothing about `\aref{}`, so my naive question is, could redefining `\ref` to `\aref` solve the problem?

Thanks!

tamas-ferenci:
Thank you! Unfortunately this redefinition wouldn't really solve the problem, as we *might* need `\ref` as well. (`aref` is only need when there is a definite article in the reference text.) If possible, the best would be to introduce `\@aref` identically to `\@ref`, with the former translating to `aref` in LaTeX, the latter to `ref`. Of course, HTML output also needs its own logic, similar to https://github.com/dmurdoch/magyarbookdown/blob/master/R/replacements.R .

yihui:
@tamas-ferenci Thanks for the explanation! We'll think more about it.

tamas-ferenci:
Dear @yihui , I was just wondering if there are any updates on this issue. Thank you very much!

yihui:
Sorry, there aren't, but I'm okay with merging this PR (the Git conflicts need to be resolved first).

tamas-ferenci:
That would be great, if you could do itm thank you in advance!

PR #691, opened 2019-03-29
added script to focus on .page-wrapper

darshanbaral:
In relation to SO question (https://stackoverflow.com/q/55379127/7128934), adds a script to focus on the main content so that the book can be navigated with the use of keyboards.


yihui:
That is not quite the correct fix, but thanks a lot anyway! I just pushed the fix via ff6cd68.

Issue #705, opened 2019-04-21
Feature request: Toggle clean_pandoc2_highlight_tags() in html_document2


atusy:
By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('bookdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/bookdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

In `html_document2`, I faced a failure in line numbering on code blocks powered by pandoc, and found this is caused by `clean_pandoc2_highlight_tags`.

I asked reasons for the cleaning on RStudio Community, and I got an answer it is probably related to GitBook (https://community.rstudio.com/t/number-lines-on-code-blocks-in-bookdown-htmld-document2-fails-because-of-bookdown-clean-pandoc2-highlight-tags/28985/2).

If this is the case, and cleaning up is not always required, I want an option to toggle the cleaning (e.g., by YAML front matter?).

This feature request enhances other packages depending on `bookdown::html_document2` such as `pagedown` and `blogdown`.
I'd be verry happy if my suggestion is accepted.

# Reproducible example

Please knit followings in `html_document` and `html_document2`.

The latter will remove the line numbering on code blocks.

````
---
output:
  html_document:
    highlight: pygments
  bookdown::html_document2:
    highlight: pygments
---
 
```{r, class.source = "numberLines lineAnchors", eval = FALSE}
iris
```
````

#  `xfun::session_info('bookdown')`

```
R version 3.5.3 (2019-03-11)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 9 (stretch), RStudio 1.2.1335

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C             
[...]

PR #706, opened 2019-04-21
feat: toggle clean_pandoc2_highlight_tags in html_document2

atusy:
I implemented a toggling option to the `clean_pandoc2_highlight_tags` in `html_document2` (Issue https://github.com/rstudio/bookdown/issues/705)

The toggling is achieved by specifying `highlight_cleaned` in a YAML front matter, which is `TRUE` in default for consistency with previous versions.

If there is any better name for this option, please tell me.

Here's a reproducible example.

 ````
---
output:
  bookdown::html_document2:
    highlight: pygments
    highlight_cleaned: false
---
 
```{r, class.source = "numberLines lineAnchors", eval = FALSE}
head(iris)
```
````

### DIFF
--- a/R/html.R
+++ b/R/html.R
@@ -110,6 +110,8 @@ tufte_html_book = function(...) {
 #'   (the i-th figure/table); if \code{FALSE}, figures/tables will be numbered
 #'   sequentially in the document from 1, 2, ..., and you cannot cross-reference
 #'   section headers in this case.
+#' @param highlight_cleaned Whether to remove the <div> tags around <pre>,
+#'   and clean up <a> on all lines in code blocks (default: `TRUE`).
 #' @inheritParams pdf_book
 #' @return An R Markdown output format object to be passed to
 #'   \code{rmarkdown::\link{render}()}.
@@ -121,7 +123,8 @@ tufte_html_book = function(...) {
 #' @references \url{https://bookdown.org/yihui/bookdown/}
 #' @export
 html_document2 = function(
-  ..., number_sections = TRUE, pandoc_args = NULL, base_format = rmarkdown::html_document
+  ..., number_sections = TRUE, pandoc_args = NULL, base_format = rmarkdown::html_document,
+  highlight_cleaned = TRUE
 ) {
   base_format = get_base_format(base_format)
   config = base_format(
@@ -134,7 +137,7 @@ html_document2 = function(
     x = restore_appendix_html(x, remove = FALSE)
     x = restore_part_html(x, remove = FALSE)
     x = resolve_refs_html(x, global = !number_sections)
-    x = clean_pandoc2_highlight_tags(x)
+    if (highlight_cleaned) x = clean_pandoc2_highlight_tags(x)
     write_utf8(x, output)
     output
   }
--- a/man/html_document2.Rd
+++ b/man/html_document2.Rd
@@ -11,7 +11,7 @@
 figures/tables/equations}
 \usage{
 html_document2(..., number_sections = TRUE, pandoc_args = NULL, 
-    base_format = rmarkdown::html_document)
+    base_format = rmarkdown::html_document, highlight_cleaned = TRUE)
 
 tufte_html2(..., number_sections = FALSE)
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by yihui** [comment]
FYI, I renamed `highlight_cleaned` to `clean_highlight_tags`.

**Comment by RLesur** [comment]
Please note that with **Pandoc 2.7.3**, the `clean_highlight_tags` argument does not toggle line numbering anymore: with `clean_highlight_tags: true` a code chunk with class `numberLines` will be numbered.

I do not interpret this behavior as a bug nor a regression. It seems to me that **Pandoc 2.7.3** makes the `clean_highlight_tags` option useless. There is no issue with that.

Refs: https://github.com/rstudio/bookdown/issues/733#issuecomment-506913254 and https://github.com/jgm/skylighting/blob/master/changelog.md#08----2019-05-27

**Comment by atusy** [comment]
Thanks.

I see that only the difference caused by  `clean_highlight_tags: true` is remaining of `<div class="sourceCode"></div>`.
This seems to be because `<a>` tags no more have the `sourceLine` class, and thus they won't be cleaned by  `clean_pandoc2_highlight_tags()`.

Yet, `true` seems to be important for better styling because Pandoc embeds following CSS to modify margins when  any chunk has class `numberLines` (see https://gist.github.com/atusy/d6d686de8982a36010c36eda94fed1a5 for full version).

``` css
div.sourceCode { margin: 1em 0; }
pre.sourceCode { margin: 0; }
```

By default, `pre` has margin based on `bootstrap.min.css`

``` css
pre {margin: 0 0 10px}
```

If we do not have to care about `clean_highlight_tags`, that's better in terms of user experience.
I think we have two options for that:

- Updating `clean_pandoc2_highlight_tags()` to only remove `<a>` not `<div>`. Though I don't know the importance of removing `<div>`
- Updating Pandoc templates to specify appropriate margins for `pre.sourceCode`.

RLesur:
About the two options described by @atusy 

- > Updating `clean_pandoc2_highlight_tags()` to only remove `<a>` not `<div>`.

  IMO, modifying the markup produced by **bookdown** is a risky option. That could have many side effects for the users and lead to regressions which are hard to resolve. See for instance, [this SO question about these divs](https://stackoverflow.com/a/48593141/6500804). 

- > Updating Pandoc templates to specify appropriate margins for `pre.sourceCode`.

  IMO, modifying CSS stylesheets is from far more flexible.

atusy:
I agree with @RLesur

yihui:
Hi @atusy, I'd like to get rid of the whole `clean_pandoc2_highlight_tags()` thing now. This hackish function was added via cd2fb76f818528eb9effd369e3a6ba558be6df68 mainly because of one thing that I didn't like about Pandoc 2.0: https://groups.google.com/d/msg/pandoc-discuss/1bwAre8fG6E/k4vM455kCQAJ Now it seems it no longer adds random IDs to lines, so `clean_pandoc2_highlight_tags()` is no longer useful to me (more importantly, I really don't like this hack myself).

What do you think?

atusy:
Nice! The  less hackish, the less vulnerable the bookdown package is to the changes in pandoc and skylighting. Let's get rid of it.

After getting rid of it, I will close or update PR #738.
I don't have to update `clean_pandoc2_highlight_tags()` and `default.html` any more.
A point is whether to modify gitbook's template or just forget about it.
What do you think?

I don't really need it because I don't use line numbers, too.
I made the PR because my friend wanted it and it was a good chance to learn inside of rmarkdown and related packages.

yihui:
I just removed the function, although I introduced another hack... https://github.com/rstudio/bookdown/commit/52b71f9575e1abd2c165781a56b2ead2ef1a7777#diff-0e9d8f9e55fa1ae0adee00b6ed6ba976R51

I haven't tried to understand #738 yet but your description seems to be very clear. Since your issues and PRs are always so clear, just let me know if you want write access to this repo someday, and I'll be happy to grant it to you :)

atusy:
Thanks for your commits and suggestions.
I'd be happy to have write access and continue contributing :smile: 
(though I guess I make PRs instead of direct commits for the most of the cases)

For #738, I closed it for once because I wanted some rethinking as commented in there.

yihui:
Okay, I just gave you write access. If you are very sure about a change, just push to the master branch. I wouldn't mind it. If you want to discuss, you can always send PRs. Thank you!

atusy:
Thanks a lot! May I add myself as `ctb` on `DESCRIPTION`?

yihui:
Of course.

atusy:
Thanks again. I did it :smiley:

Issue #741, opened 2019-07-13
Please develop bookdown::powerpoint_presentation2 output for cross-references


jooyoungseo:
As `rmarkdown::render()` now officially supports `powerpoint_presentation` output_format, would you mind developing an advanced wrapper that enables cross-references like `bookdown::powerpoint_presentation2` just as in `bookdown::word_document2`?
That would be greatly appreciated.

yihui:
Done. Many thanks to @atusy for the implementation!

PR #742, opened 2019-07-14
Add odt_document2 and powerpoint_presentation2

### LINKED ISSUE #722: it would be great to support OpenDocument (odt) file format for output
As you wrote [in the book](https://bookdown.org/yihui/bookdown/r-markdown.html) the  `rmarkdown` already supports `odt_document` output format.

It would be really great if you add support for OpenDocument (odt) file format to `bookdown` and name it `bookdown::odt_document2` for convenience.

atusy:
This PR adds `odt_document2` and `powerpoint_presentation2` and closes #722 #741 
Though I'm not sure `powerpoint_presentation2` is a scope of bookdown, bookdown is the only place to implement it.

## odt

I rendered `inst/examples/index.Rmd`.
The result seems to be fine.

[bookdown.zip](https://github.com/rstudio/bookdown/files/3389363/bookdown.zip)

Exceptionally, when math expressions appeear right after the code blocks, the math expressions appear within code blocks.
However, this is not a problem of bookdown because the same problems occur in `rmarkdown::odt_document`.

image

## pptx

Cross reference seems to be working fine

image

````
---
output: bookdown::powerpoint_presentation2
---

# Cross-reference

```{r iris, fig.cap="iris"}
plot(iris)
```

`\@ref(fig:iris)` becomes \@ref(fig:iris)
````

### CODE REVIEW

**Comment by atusy** [comment]
May be it's better to factor out duplicated codes into a new function (`office_document2_base`?) from `word_document2`, `powerpoint_presentation2`, and `odt_document2`?

**Comment by yihui** [comment]
Yes, that is exactly what I was thinking. Please do factor out the code instead of repeating it in three functions. Thank you!

**Comment by atusy** [comment]
Done it by implementing `office_document2_base`.
It is not exported currently, but might be useful for users in case they want some formats to be transformed to `bookdown`'s format.

As rewriting extremely simplfied `odt.R` and `powerpoint.R`, I integrated them into `word.R`.
Maybe we should renaming `word.R` to `office.R` or somethingl like that.

File: R/word.R [hunk 1]
```diff
@@ -1,24 +1,38 @@
-#' @rdname html_document2
-#' @export
-word_document2 = function(fig_caption = TRUE, md_extensions = NULL, pandoc_args = NULL, ...) {
-  from = rmarkdown::from_rmarkdown(fig_caption, md_extensions)
+office_document2_base = function(base_format) {
```
yihui: This is good, but I slightly prefer this style:

```r
office_document2_base = function(
  fig_caption = TRUE, md_extensions = NULL, pandoc_args = NULL, ..., base_format
) {

}
```

It will be more consistent with other functions like
https://github.com/rstudio/bookdown/blob/9516a435773fc55d822d10012101405af54f6274/R/html.R#L123-L125

What's more, we can export this function and allow users to customize all arguments of `base_format()`.

Actually, let's rename it to `markdown_document2()` (because it is not only useful to Office documents) and export it. You will also need to document it. It may take you some time to explain what this function really does.

atusy: Got it. So updated `word_document2` will be

``` r
word_document2 = function(
  ..., fig_caption = TRUE, md_extensions = NULL, pandoc_args = NULL
) {
  markdown_document2(
    fig_caption = fig_caption, md_extensions = md_extensions, pandoc_args = pandoc_args,
    ..., base_format = rmarkdown::word_document
  )
}
```

or more simply like `html_book`,

``` r
word_document2 = function(...) markdown_document2(..., base_format = rmarkdown::word_document)
```

Which do you prefer?

Do I document `markdown_document2` in `@rdname html_document2` or in an individual file?

yihui: 1. The latter.

2. You can do `@rdname html_document2`.

atusy: Thanks and I did it :)

atusy:
I confirmed `markdown_document2` well works with other formats (e.g., `prettydoc::html_pretty`.
Note that hyperlinks to referred tables and others are not working.

````
---
output: 
  bookdown::markdown_document2:
    base_format: prettydoc::html_pretty
---

# Table \@ref(tab:iris)

```{r iris}
knitr::kable(head(iris), caption = "iris")
```

# Table \@ref(tab:mtcars)

```{r mtcars}
knitr::kable(head(mtcars), caption = "mtcars")
```
````

**Comment by jooyoungseo** [comment]
> I confirmed `markdown_document2` well works with other formats (e.g., `prettydoc::html_pretty`.
> Note that hyperlinks to referred tables and others are not working.
> 
> ```
> ---
> output: 
>   bookdown::markdown_document2:
>     base_format: prettydoc::html_pretty
> ---
> 
> # Table \@ref(tab:iris)
> 
> ```{r iris}
> knitr::kable(head(iris), caption = "iris")
> ```
> 
> # Table \@ref(tab:mtcars)
> 
> ```{r mtcars}
> knitr::kable(head(mtcars), caption = "mtcars")
> ```
> ```

@atusy, `bookdown::markdown_document2` does not seem to work with [redoc](https://github.com/noamross/redoc) package.

Cross-references are broken in the rendered results. Would you mind testing it with the following?:

````
---
output: 
  bookdown::markdown_document2:
    base_format: redoc::redoc
---

Table \@ref(tab:iris)

```{r iris}
knitr::kable(head(iris), caption = "iris")
```

Table \@ref(tab:mtcars)

```{r mtcars}
knitr::kable(head(mtcars), caption = "mtcars")
```

Figure \@ref(fig:histogram)

```{r histogram}
hist(airquality$Ozone, col="blue")
```
````

**Comment by atusy** [comment]
As far as I tried, broken cross-references occur only at `Table \@ref(tab:iris)`.
If `keep_md: true` is specified, you see the line is converted to

```md
Table \<span class="redoc" id="redoc-citation-1"><span class="redoc" id="redoc-citation-2"><span class="redoc" id="redoc-citation-3">@ref</span></span></span>(tab:iris)
```

I think this is the reason and is caused by `redoc` package.
Maybe @noamross knows what is happening.

Besides, `Figure \@ref(fig:histogram)` is converted to `Figure ??` in docx.
This is because you did not specify `fig.cap` for the histogram chunk.

BTW, I recommend opening issue when you see strange behavior instead of making comments on closed PR.
Then, more people notices the issue and possibly solve it more quickly.
(This is just my opinion)

noamross:
This is most likely caused by the **redoc** package as I haven't yet done the work to make sure it's post-processing is compatible with these formats.  Feel free to open an issue over there.

PR #745, opened 2019-07-17
fix 2 small XHTML output bugs

larskotthoff:
- `link` tags need to be closed in XHTML (see https://www.w3schools.com/tags/tag_link.asp)
- `&` isn't a valid XML character on its own


larskotthoff:
I'm not sure to be honest, that's why I bypassed the issue. It renders without errors if you use `&amp;`, but that's what literally appears in the Javascript in the page source, so I don't think it would work.

PR #758, opened 2019-08-16
added github button doc

bblodfon:


inst/examples/03-formats.Rmd
```diff
@@ -103,6 +103,7 @@ bookdown::gitbook:
       size: 2
     sharing:
       facebook: yes
+      github: yes
```
yihui: The default is `no`:

```suggestion
      github: no
```

bblodfon: Yes, true! I just took it from my use-case where I wanted just this button and nothing else

yihui: Would you change it to `no`?

bblodfon: Sorry. new to pull requests. Is everything ok now?

PR #768, opened 2019-08-31
rtf_document2 added

jooyoungseo:
- Added new output format `rtf_document2` based on rmarkdown::rtf_document for cross-referencing capabilities.


atusy:
@Yihui Thanks, I agree. I will keep that in my mind.

jooyoungseo:
I am very sorry for this trouble @yihui and @atusy.
Since I don't (cannot) use RStudio due to its accessibility issue, I used `devtools::check()` manually in the `bookdown` package directory after change was made.

If I use `Rd2roxygen::rab()` instead of `devtools::check()`, would it be suitable for PRs moving forward, @yihui?
Please advise me, and I massively apologize for this confusion again.

yihui:
@jooyoungseo No worries and no need to apologize at all! I completely understand your situation. It is partly my fault because I used a "non-standard" way for roxygenizing. That is why I said I might give it up in the future to avoid further confusion. You are not the first person who has been hit by this problem. You just do whatever convenient to you, and I was asking @atusy to help amend the PRs in case this happens again in the future. Thanks for all your contributions!

Issue #754, opened 2019-08-13
Problematic assessment of CSL/pandoc-citeproc in the docs


njbart:
I feel the opening passages in <https://bookdown.org/yihui/bookdown/citations.html> do not do justice to CSL in general, and pandoc-citeproc in particular.

> Although Pandoc supports multiple ways of writing citations, we recommend you to use BibTeX databases because they work best with LaTeX/PDF output.

To my mind, this is not a fair assessment. Neither are BibTeX databases inherently “better”, nor do BibTeX or biber/BibLaTeX (the programs) necessarily do a better job processing these than pandoc-citeproc does. Sure, pandoc-citeproc has to convert BibTeX or BibLaTeX data to CSL-JSON first, but it takes great care to make this process as lossless as possible. On the other hand, if bibliographic data are already available in the CSL-JSON format (e.g., from Zotero, or similar CSL-based reference managers), there is no reason at all to convert these to BibTeX when using pandoc-citeproc.

Also, I would argue that BibTeX (the program) itself is pretty much obsolete, and its data model severely limited by modern standards. (Even <https://www.ctan.org/pkg/bibtex> concedes that its “facilities rapidly run out as one moves away from simple ASCII […] For more flexibility, the user is urged to consider using biber with BibLaTeX […]“)

BibTeX’s more modern successor, BibLaTeX, arguably is on a par with CSL/pandoc-citeproc in terms of their data models and processing capabilities, at least as far as the average user wishing or needing to follow a particular style guide is concerned. (I’ll readily admit that BibLaTeX allows a lot more fine-tuning, provided one has mastered that particular art. However, BibLaTeX/BibTeX is limited to one single output format, PDF, whereas no such limitations exist for CSL/pandoc-citeproc. In addition, CSL provides far more well-maintained styles implementing the recommendations of recognised style guides, including popular ones, such as Chicago, APA, MLA, etc., and more niche variants for specific journals or university departments than BibLaTeX does.)

> Pandoc can process other types of bibliography databases with the utility pandoc-citeproc (https://github.com/jgm/pandoc-citeproc), but it may not render certain bibliography items correctly (especially in case of multiple authors per item).

I’d kindly ask to see the evidence for this rather sweeping claim.

[...]

PR #772, opened 2019-09-08
Update 02-components.Rmd

njbart:
See https://github.com/rstudio/bookdown/issues/754

### DIFF
--- a/inst/examples/02-components.Rmd
+++ b/inst/examples/02-components.Rmd
@@ -691,196 +691,28 @@ For all custom blocks based on the `block` or `block2` engine, there is one chun
 
 ## Citations {#citations}
 
-rmarkdown and bookdown offer two different methods for managing
-citations and bibliographic references in a document.
+Pandoc offers two methods for managing citations and bibliographic references in a document.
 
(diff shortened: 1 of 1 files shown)


atusy:
Thanks for the insightful documentation.
OTH, I feel the new "Citation" section seems to be relatively heavy compared to the other sections.
Just like the "Markdown syntax" section, we can let readers find comprehensive introduction in the Pandoc's manual.

In particular, wouldn't it be better to explain mainly about CSL and pandoc-citeproc?
This is because this is the only way to support variety of output formats including PDF.

njbart:
Just re-read my text. Based on my experience trying to explain this stuff to novices I continue to feel that essentially all of this information is required to sort out the precise relations between pandoc-citeproc, bibtex and biblatex. Still, if you want to rewrite my text, fine.

Nonetheless, the current version of https://bookdown.org/yihui/bookdown/citations.html is still both *factually inaccurate and needlessly disparaging* (“pandoc-citeproc … may not render certain bibliography items correctly”), and *incomplete* (it does not even mention biblatex). These issues, at the very least, should be fixed as a matter of urgency.

PR #776, opened 2019-09-19
[gitbook] Copy to clipboard buttons for code blocks

### LINKED ISSUE #775: Feature request: a copy-to-clipboard button for code
pkgdown has a very useful new feature: a copy-to-clipboard button for code. You can see this if you hover over the code on [this page](https://tidyr.tidyverse.org/articles/pivot.html).

This would be a very useful feature to bring to bookdown, for the output formats that support interactivity.

@RLesur

RLesur:
The aim of this PR is to resolve #775. 

I propose the following implementation:

- use [clipboard.js](https://github.com/zenorocha/clipboard.js/): this is a maintained lightweight JS library for copying to clipboard. It is already used by **pkgdown**. There exists a [gitbook `copy-code` plugin](https://github.com/WebEngage/gitbook-plugin-copy-code-button) but it seems to be unmaintained. @yihui, tell me if you agree to integrate **clipboard.js** in **bookdown** or if you would prefer to use the `copy-code` plugin.

- a very basic styling for copy-to-clipboard buttons. @behrman, contrary to **pkgdown** there is no tooltip. For tooltips, we should add another library in **bookdown** and I feel that the extra benefit of tooltips would be very small.


behrman:
I've tested the feature branch, and it seems to work fine. We do get a tooltip if we hover over the copy-to-clipboard icon for a sufficient amount of time. I think this is fine.

RLesur:
@behrman For the tooltip, we do get the user agent button tooltip. Its design and styling may vary from one browser to another. 

The main difference with **pkgdown** is that the tooltip content is not updated after a succesfull copying (the user sees `Copied!` with **pkgdown**). 

There is another difference: in case of failure, there is a specific message in the **pkdown** tooltip, see https://github.com/r-lib/pkgdown/blob/d01672efa3baa80fd7ab238b7143080b0daf49a4/inst/assets/pkgdown.js#L108-L110
With the current PR, the copying action can silently fail.

behrman:
@RLesur You've done a fine job with the design of the tooltip messages for **pkgdown**. There may be some advantage to have consistent behavior between that and **bookdown**, though I don't know the downsides of doing this. I think the current behavior is fine.

PR #777, opened 2019-09-23
render_book supports self contained false for html_document2 and its extensions

atusy:
Currently, `render_book` moves all the dependencies to the `_bookdown_files` directory when rendering `html_document2` or its extensions.
However, the output HTML file is left in the working directory.
This is causing the output from `html_document2` and its extensions unable to find the dependencies.

This PR fixes this problem by suppress moving dependencies to the `_bookdown_files` directory.
Another possible approach is to move result HTML file and dependencies to `_book` directory, and rename the result HTML file to `index.html`.
I chose the former solution for backward compatibility as current `html_document2` generates an HTML file in the working directory.

In order to compare the results from current version and this PR,

1. Create Book Project using bookdown
2. Add following lines to `_output.yml`
  ```
  bookdown::html_document2:
    self_contained: false
  pagedown::html_paged:
    self_contained: false
  ```
3. Build book with `bookdown::html_document2` or `pagedown::html_paged`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 - Added copy to clipboard buttons to code blocks in the `gitbook` output format (thanks, @behrman #775, @RLesur #776).
 
+- `render_book` supports `self_contained: false` for `html_document2` and its extensions such as `pagedown::html_paged` (thanks, @atusy, #777)
+
 ## BUG FIXES
 
 - Images specified in `toc: before:` of the `gitbook` format are not copied to the output directory (thanks, @dcossyleon, #763).
--- a/R/html.R
+++ b/R/html.R
@@ -141,6 +141,7 @@ html_document2 = function(
     output
   }
   config$bookdown_output_format = 'html'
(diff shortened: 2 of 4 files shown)


atusy:
I see, and I have to explain a bit more.
With `Build Book` button on RStudio, RStudio actually runs `rmarkdown::render_site`, and `bookdown::render_book` seems to be called by `render_site`.

```r
rmarkdown::render_site(output_format = 'pagedown::html_paged', encoding = 'UTF-8')
```

Does it indicate `render_site` is the one to be fixed rather than `render_book` so that `render_site` properly calls `render` or `render_book` dynamically?

atusy:
Or RStudio is the one to be fixed so that `render` should be called instead of `render_site` for `html_document2`?

atusy:
I'm totally wrong. `render_book` is called because of `site: bookdown::bookdown_site` in the YAML front matter.
I have to look for appropriate site generator.
Thanks!

PR #778, opened 2019-09-25
Fix for caption text that ends in whitespace for Word/EPub documents

jakelever:
When using (ref:figureName) to put the caption outside the figure block, captions for figures won't render if they end in a space for Word/ePub documents. 

There is currently a regex looking for the figure captions but can't end with a space. This change allows for whitespace and removes it from the end of a caption.


jakelever:
Understood. I got a little confused because it behaves differently between HTML/PDF (space allowed) and Word/EPub (space not allowed).

Issue #744, opened 2019-07-17
Please allow users to define prefix for Appendix section in UI like Chapter


jooyoungseo:
Currently, it does not seem that users can define prefix used for "Appendix" in UI like chapter prefix.
If you click any appendix link created by bookdown, the title bar starts with an alphabet like A or B (not Appendix A or Appendix B).
For example, open [this link](https://bookdown.org/yihui/bookdown/software-tools.html) and the title bar is not prefixed with "Appendix ".
Would there be any way to achieve this goal?

yihui:
Sounds like a reasonable feature request, but I don't have time for it. I'm hoping someone else could help here...

WerthPADOH:
I'll work on this, if that's alright.

WerthPADOH:
Made a PR: #783.  I also added an option to modify the counters that come before appendix titles.  The default counters are still like A, B.1, B.2, etc.

yihui:
I have merged the PR. Please test with `remotes::install_github('rstudio/bookdown')`. The documentation is at https://bookdown.org/yihui/bookdown/internationalization.html (search for `appendix_name` there). Many thanks to @WerthPADOH for the work and @jooyoungseo for the request!

PR #783, opened 2019-10-08
Appendix prefix and counter config

WerthPADOH:
PR for #744
Add two config options for `_bookdown.yml` under `language:ui`:
- `appendix_name`: string to insert before appendix titles (default `''`)
- `appendix_counters`: list of strings to use for counters
    (default uppercase alphabet)


yihui:
FYI I decided to drop the `appendix_counters` option for the sake of simplicity (to make it easier to fix bugs like #827). This means appendices can only be numbered by A-Z.

PR #798, opened 2019-11-03
handle change in footnotes from pandoc 2.7.3

### LINKED ISSUE #793: Footnotes appear on the last html page instead of one they are placed on.
While creating my own gitbook as site (using RStudio's *Knit to gitbook*) I realized that footnotes end up on the very last HTML page of the whole site. I tried to use `split_bib: yes` explicitly (as far as I know it is default behaviour in HTML documents) but it didn't work. Thinking it has to do with my code and the book itself I tried using example provided for oxforddown: https://github.com/ulyngs/oxforddown. For example https://ulyngs.github.io/oxforddown/rmd-basics.html#footnotes uses footnotes and they appear on the same page. When I use *Knit to gitbook* in RStudio all footnotes end up in the end in the *References* section (which doesn't appear in the oxforddown version hosted online).

Thus it is either problem with my system or with bookdown. Hope someone can reproduce it.

Checked with newest R, RStudio and updated packages. 

Output of `xfun::session_info('bookdown')` below:

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18362), RStudio 1.2.5001

Locale:
  LC_COLLATE=Polish_Poland.1250  LC_CTYPE=Polish_Poland.1250    LC_MONETARY=Polish_Poland.1250
  LC_NUMERIC=C                   LC_TIME=Polish_Poland.1250    

Package version:
  base64enc_0.1.3 bookdown_0.14.1 digest_0.6.22   evaluate_0.14   glue_1.3.1      graphics_3.6.1  grDevices_3.6.1
  highr_0.8       htmltools_0.4.0 jsonlite_1.6    knitr_1.25      magrittr_1.5    markdown_1.1    methods_3.6.1  
  mime_0.7        Rcpp_1.0.2      rlang_0.4.1     rmarkdown_1.16  stats_3.6.1     stringi_1.4.3   stringr_1.4.0  
  tinytex_0.17    tools_3.6.1     utils_3.6.1     xfun_0.10       yaml_2.2.0     
```
---

- [x] I have fully read the issue guide at https://yihui.name/issue/.

cderv:
This fixes #793 

There was a change in pandoc where  the ⤶︎ was previously a single unicode `\u2936`, and now it is encoded in two unicode character `\u2936\uFE0E`. That breaks the regex that identify footnootes in html code. In regex, a dot `.` will only match one unicode, not two even if it is one. 
See example: 

``` r
grep(">.</a>", enc2utf8(">\u2936</a>"), value = TRUE)
#> [1] "><U+2936></a>"
grep(">.</a>", enc2utf8(">\u2936\uFE0E</a>"), value = TRUE)
#> character(0)
grep(">.+</a>", enc2utf8(">\u2936\uFE0E</a>"), value = TRUE)
#> [1] "><U+2936><U+FE0E></a>"
```

This is an easy fix to use `.+` instead of `.` - I used that for now.

This can also be handled using purl regex and special pattern `\X`. See [here](https://www.regular-expressions.info/unicode.html)
``` r
grep(">\\X</a>", enc2utf8(">\u2936</a>"), value = TRUE, perl = TRUE)
#> [1] "><U+2936></a>"
grep(">\\X</a>", enc2utf8(">\u2936\uFE0E</a>"), value = TRUE, perl = TRUE)
#> [1] "><U+2936><U+FE0E></a>"
```

Or directly using unicode pattern match.

``` r
grep(">\\x{2936}</a>", enc2utf8(">\u2936</a>"), value = TRUE, perl = TRUE)
#> [1] "><U+2936></a>"
grep(">\\x{2936}\\x{FE0E}</a>", enc2utf8(">\u2936\uFE0E</a>"), value = TRUE, perl = TRUE)
#> [1] "><U+2936><U+FE0E></a>"
```

@yihui, is the quick fix good enough for you or should we go with exact match ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## BUG FIXES
 
+- When using pandoc 2.7.3 or later, footnotes are now placed again at the end of each chapter. (#798, thanks @cderv) 
+
 - gitbook toolbar is not missing any more when rendering books with Pandoc 2.x and using `self_contained = TRUE` (thanks, @Pindar777, @RLesur, @cderv, #739).
 
 # CHANGES IN bookdown VERSION 0.14
--- a/R/html.R
+++ b/R/html.R
@@ -876,7 +876,7 @@ parse_footnotes = function(x) {
   j = which(x == '</div>')
   j = min(j[j > i])
   n = length(x)
-  r = '<li id="fn([0-9]+)"><p>.+?<a href="#fnref\\1"[^>]*?>.</a></p></li>'
+  r = '<li id="fn([0-9]+)"><p>.+?<a href="#fnref\\1"[^>]*?>.+</a></p></li>'
   s = paste(x[i:n], collapse = '\n')
   items = unlist(regmatches(s, gregexpr(r, s)))
   list(items = setNames(items, gsub(r, 'fn\\1', items)), range = i:j)


Pindar777:
@cderv I just noted the issue and installed the devtools bookdown version. Now all footnotes are printed the first time a reference appears. I doubt that this is the solution of the fix. Is there a way to install this fix via remotes? 
Thanks

cderv:
@Pindar777 I am not sure to follow. Can you open a new issue with your example so that I can try ? 
Playing with regex is not an easy task and maybe this is not enough to a `+`

Pindar777:
@cderv Thanks for your fasrt response.
I could reproduce the error with this simple demo where I tested some user definded blocks.

[bookdown-demo-footnote_blocks.zip](https://github.com/rstudio/bookdown/files/3804520/bookdown-demo-footnote_blocks.zip)

cderv:
Ok; there is indeed a second issue where the footnote are not placed correctly after being found. Can you open a new issue to track that correctly ? 

I will dig into that later today. 

Thanks a lot for the test and the reports 🙏

PR #802, opened 2019-11-04
Update sharing urls

cderv:
This closes #797 

As explained in https://github.com/rstudio/bookdown/issues/797#issuecomment-549181261 there was changes in sharing url. 

This PR: 

* removes google+ as it does not exist anymore. 
* updates sharing link for twitter and facebook
* changes icon for instapaper as `fa-instapaper` does not exists.

`fa-italic`
image

Instapaper icone 
image

## About google removal

Google sharing is removed from example and default value. 
A new function `check_gb_config` is here to check configuration and change user's config with a warning to disable some option. (here remove google from `sharing` list or `all` vector

I think this is enough to not break older bookdown where `google` would have been present in configuration. 

We could also leave as is and the sharing link would redirect to google+ removal page
https://plus.google.com/

See for example https://bookdown.org/yihui/bookdown/ and sharing link to google plus.

PR #813, opened 2019-11-12
check for google sharing handles disabled sharing too

### LINKED ISSUE #812: bookdown fails with "sharing: false" after "check_gb_config()" check added
---

- [X] I have fully read the issue guide at https://yihui.name/issue/.
- [X] I have provided the necessary information about my issue.
- [X] I have learned the Github Markdown syntax, and formatted my issue correctly.

---

After the new update based on commit 32a77395a32a0b6989a6df1546ef5b28fcf69975 and the `check_gb_config()` was added, using `sharing: false` or `sharing: no` in `_output.yml` leads to error: `Error in config[["sharing"]][["google"]] : subscript out of bounds`. This behaviour did not occur before. This is due, I assume, to the `check_gb_config()` function assuming the config "sharing" option is a list with a variable called "google"... when in this case it is not.

Is this a new expected behaviour or no?

cderv:
This closes #812 and is a bugfix to a regression following #802 

While adding backward compatibility, `sharing = FALSE`( `sharing: no` or `sharing: false` and `sharing: null` were not handled correctly. 

This is now fixed. 

I also added some tests using testit to prevent this  - I tested all combination of configuration for sharing I could think of.


cderv:
@yihui I think this is good now. Sorry this has happened just after the CRAN release 😟

cderv:
I went back and modified `plugin-sharing.js` only

It was in fact already dealt with for sharing buttons creation with `if (!opts[sideId]) return;`, it is now handled also for menu creation too. If a configuration in sharing is not among facebook, github, twitter, linkedin, weibo, instapaper or vk, it will be skipped and do nothing. 

It is way easier to deal with Google+ removal and any other non supported sharing config.

PR #820, opened 2019-11-17
TOC collapse by section works with accents in chapter html filename

### LINKED ISSUE #819: Chapter title containing accents does not unfold sections in ToC
Hello.

It is easy to reproduce this bug. Start with the *bookdown-demo* template. Then, create a new chapter including an accent in any of the words in the chapter title (for instance, if you write in Spanish, French, etc.). Include 2 section headers in the new chapter, for testing purposes.

```
# Título de capítulo

## Entrada uno

## Entrada dos
```
Use in `_output.yml` the following options to keep chapters folded in ToC at first load:

```
bookdown::gitbook:
  config:
    toc:
      collapse: section
      ... (rest of options)
```

Testing the ToC on the built website, when you click on the chapter entry in the ToC  the link is not hightlighted and the list of sections does not unfold. I think the link gets broken, somehow (not for rendering the chapter on the main section, but for the ToC highlighting functionality.

BTW, I have just checked that, even if the chapter title contains accents, if you add a specific link like in the example below, everything works again (list of sections unfolds, chapter name keeps highlighted, etc.):

```
# Título de capítulo con acentos {#mi-titulo}
```
It will not affect most users writing in English, but is quite annoying for users in other languages with accents (I have spent several hours with this bug, until I figured out what was going on).

Session info:

```
> xfun::session_info('rmarkdown')
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17763), RStudio 1.2.5019

Locale:
  LC_COLLATE=Spanish_Spain.1252  LC_CTYPE=Spanish_Spain.1252    LC_MONETARY=Spanish_Spain.1252
  LC_NUMERIC=C                   LC_TIME=Spanish_Spain.1252    

Package version:
  base64enc_0.1.3 digest_0.6.22   evaluate_0.14   glue_1.3.1      graphics_3.6.1 
  grDevices_3.6.1 highr_0.8       htmltools_0.4.0 jsonlite_1.6    knitr_1.26     
  magrittr_1.5    markdown_1.1    methods_3.6.1   mime_0.7        Rcpp_1.0.3     
  rlang_0.4.1     rmarkdown_1.17  stats_3.6.1     stringi_1.4.3   stringr_1.4.0  
  tinytex_0.17    tools_3.6.1     utils_3.6.1     xfun_0.11       yaml_2.2.0     

Pandoc version: 2.7.2
```
---

- [x] I have fully read the issue guide at https://yihui.name/issue/.

cderv:
This will fix #819 

When there is some special characters (like accentuated ones) in chapter titles, the default `split_by: chapter` will create several html file with the a name like the chapter title. 

Example from bookdown demo: 
* `litérature.html` in French if you put `# Litérature` in file `02-literature.Rmd`

In this case, as described in #819, there is an issue in the TOC. This is because 
https://github.com/rstudio/bookdown/blob/cc9dcecbfe510ea01eefc8d8651662b9164fb25b/inst/resources/gitbook/js/plugin-bookdown.js#L98-L99
will get you url encode filename
```js
var href = window.location.pathname;
href = href.substr(href.lastIndexOf('/') + 1);
href
> "lit%C3%A9rature.html"
```
All the JS logic afterward is broken because it will compare this href with some attributes and even use it as attributes in some Jquery selection. it should be `href = litérature.html` instead.
Example: 
* This won't select the `li` corresponding to `litérature.html`
https://github.com/rstudio/bookdown/blob/cc9dcecbfe510ea01eefc8d8651662b9164fb25b/inst/resources/gitbook/js/plugin-bookdown.js#L101
* This won't be equal to true where it should
https://github.com/rstudio/bookdown/blob/cc9dcecbfe510ea01eefc8d8651662b9164fb25b/inst/resources/gitbook/js/plugin-bookdown.js#L131

The easy solution in JS I suggest here is to consider that `window.location.pathname` should be url-decoded before processing. 

The other solution is more deeply in bookdown code: **Bookdown could consider that there should not be any special characters in filenames**. This would mean that `split_chapter` should be modified to not create `litérature.html`  but `literature.html` even in accent in the chapter title. 
However, I think this could lead to more breakage as it would require to handle filename reference without any special character different from chapter name in toc and body with accentuated character. It seems more complicated. 

This modification is not easy to test. I just played with bookdown demo and special title.

@yihui I let you look into this and see for yourself if the JS fix seems enough or there should be a better logic  used for accentuated chapter title in bookdown.

### DIFF
--- a/inst/resources/gitbook/js/plugin-bookdown.js
+++ b/inst/resources/gitbook/js/plugin-bookdown.js
@@ -28,7 +28,7 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
       }
     });
 
-        // add the View button (file view on Github)
+    // add the View button (file view on Github)
     var view = config.view;
     if (view && view.link) gitbook.toolbar.createButton({
       icon: 'fa fa-eye',
@@ -97,6 +97,8 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
     // highlight the current section in TOC
     var href = window.location.pathname;
     href = href.substr(href.lastIndexOf('/') + 1);
+    // accentuated characters need to be decoded (#819)
+    href = decodeURI(href);
     if (href === '') href = 'index.html';
     var li = $('a[href^="' + href + location.hash + '"]').parent('li.chapter').first();
     var summary = $('ul.summary'), chaps = summary.find('li.chapter');
--- a/inst/resources/gitbook/js/plugin-bookdown.js
+++ b/inst/resources/gitbook/js/plugin-bookdown.js
@@ -98,7 +98,7 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
     var href = window.location.pathname;
     href = href.substr(href.lastIndexOf('/') + 1);
     // accentuated characters need to be decoded (#819)
-    href = decodeURI(href);
+    href = decodeURIComponent(href);
     if (href === '') href = 'index.html';
     var li = $('a[href^="' + href + location.hash + '"]').parent('li.chapter').first();
     var summary = $('ul.summary'), chaps = summary.find('li.chapter');
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 - For the `gitbook` output format, disabling the `sharing` menu or buttons works again (thanks, @lwjohnst86, #812).
 
+- For the `gitbook` output format, toc collapsed by section now works with accents in chapter titles (thanks, @glimmerphoenix, @cderv, #819) 
+
 # CHANGES IN bookdown VERSION 0.15
 
 ## BUG FIXES


inst/resources/gitbook/js/plugin-bookdown.js
```diff
@@ -97,6 +97,8 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
     // highlight the current section in TOC
     var href = window.location.pathname;
     href = href.substr(href.lastIndexOf('/') + 1);
+    // accentuated characters need to be decoded (#819)
+    href = decodeURI(href);
```
yihui: I guess you should call `decodeURIComponent()` here.

cderv: I hesitated between both and it seemed special characters URI separators (; / ? : @ & = + $ , #) seems to already be handled. I tested with a title `litérature ?` and `?` was not put inside the filename, so not encoded. 

However, it also will work as expected with `decodeURIComponent` so it is ok to use it instead.

PR #693, opened 2019-04-02
Updated Jquery to 3.3.1

mterente:
As Jquery < 3.0.0 is vulnerable to XSS - see https://www.cvedetails.com/cve/CVE-2015-9251/, I'd like to suggest you to update it to the latest version, with this PR.

PR #882, opened 2020-04-30
Security patch: Updates jquery version with 3.5.0

cooknl:
I received a GitHub security alert with a suggested remediation to update jquery to 3.5.0.

This is the only place in bookdown that I could find an opportunity to do so.


cooknl:
Would you like me to add and commit?

cooknl:
I have replaced jquery.min.js v2.2.3 with v.3.5.0. All tests have passed.

I have done a global search in the repo for the string "jquery"

Two questions:

1. In /bookdown/inst/resources/gitbook/js/app.js there is a comment that mentions "jQuery JavaScript Library v2.1.4". Does this comment need to mention the version number? Especially since it mentions v2.1.4 and the current version is v2.2.3? (notably app.min.js doesn't have any comments and does not appear to have any particular version dependency)
1.  In /bookdown/inst/resources/AUTHORS jQuery contributors are listed. Does this need to be updated for jQuery v.3.5.0? If so, what is the truth source for this listing?

yihui:
This has been done in #693. To answer your two questions (if you still need the answers):

1. No, we don't need to mention the version number (I've removed it in #693); app.js just uses whatever version of jQuery that is currently loaded.
2. Yes, the author info needs to be updated (and done in #693).

Thank you very much!

PR #886, opened 2020-05-11
correcly get source files when rmd_files and rmd_subdir are provided

### LINKED ISSUE #885: Build w/ rmd_files: ["index.Rmd"]<cr> rmd_subdir: "chapters" includes ./_Notes.Rmd
I've attached a RStudio project that reproduces this error. Here is a brief summary:

_bookdown.yml has these two lines:

```
rmd_files: ["index.Rmd"]
rmd_subdir: ["chapters"]
```

There's also a file in the project's root directory, ./_Notes.Rmd. Now, as I understand how the rmd_ lines above are supposed to work during a build is to read index.Rmd and then recursively include everything in ./chapters/. But the build results include the contents of ./_Notes.Rmd. They shouldn't.
[BookdownTest.zip](https://github.com/rstudio/bookdown/files/4601934/BookdownTest.zip)

This is my very first new issue on github. I didn't notice before, but the zip file of my reproducible example was added twice. I don't think this should be a problem, and I'm afraid of damaging the issue bundle by deleting it here. So it appears twice. 

```
xfun::session_info()
R version 4.0.0 (2020-04-24)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6, RStudio 1.2.5042

Locale: en_US.UTF-8 / en_US.UTF-8 / en_US.UTF-8 / C / en_US.UTF-8 / en_US.UTF-8

Package version:
  base64enc_0.1.3 bookdown_0.18   compiler_4.0.0  digest_0.6.25   evaluate_0.14   glue_1.4.0     
  graphics_4.0.0  grDevices_4.0.0 highr_0.8       htmltools_0.4.0 jsonlite_1.6.1  knitr_1.28     
  magrittr_1.5    markdown_1.1    methods_4.0.0   mime_0.9        packrat_0.5.0   Rcpp_1.0.4.6   
  rlang_0.4.6     rmarkdown_2.1   stats_4.0.0     stringi_1.4.6   stringr_1.4.0   tinytex_0.22   
  tools_4.0.0     utils_4.0.0     xfun_0.13       yaml_2.2.1   
```

---

[BookdownTest.zip](https://github.com/rstudio/bookdown/files/4601887/BookdownTest.zip)

[BookdownTest.zip](https://github.com/rstudio/bookdown/files/4601941/BookdownTest.zip)

cderv:
This fixes #885 and follows a series of features addition for `source_files` (#561, #601) in the past. 

There was still an issue when `rmd_files` is provided with `rmd_subdir` due to the following behavior: 
when `rmd_subdir` was provided, all the files from root dir and subdirs were selected. `rmd_files` was not able to filter out the desired root files. 

This PR will identify subdir files apart from root files if `rmd_subdir` is provided, then if `rmd_files` is provided also, it will concatenate `rmd_files` with `rmd_subdir` content only, discarding any other root files content. 

This PR also add tests so that all the desired behavior are checked: 
* by default, only root files are selected apart from ignored files (starting with `_`)
* `rmd_files` allows to select the root files and change default behavior
* `rmd_subdir` allows to add some subdirs content - by default root files 
* `rmd_subdir = TRUE` will get all the subdirs Rmd content. 
* if `rmd_files` is not provided, the files starting with an underscore in subdirectories will also be ignored.
* `rmd_files` allows to select the root files selected, in addition to `rmd_subdir`, adding those with the subdir content.

However, in this last case, there is one behavior which is currently happening and not sure it is desired: if `rmd_files` AND `rmd_subdir` are provided, the files starting with `_` inside sub directories will not be ignored anymore.  **Is this the desired behavior?**

I wonder if
* `_*.Rmd` files should be always ignored in bookdown. Would be a change in default, but you just need to remove the `_` and use `rmd_files` if you want to include it in some cases
* if it should be ignored by default in subdir content in all case. 

What do you think ?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,11 @@
 
 - Text references do not work for `theorem` environments (thanks, @ssp3nc3r, rstudio/tufte#75).
 
-- when `rmd_subdir` and `rmd_files` are provided, only the `rmd_files` are now selected in addition to the `rmd_subdir` content. (thanks, @Gnossos, @cderv, #885).
+- When both `rmd_subdir` and `rmd_files` are provided in the config file `_bookdown.yml`, only the files specified in `rmd_files` are now selected in addition to files under `rmd_subdir`. In the previous version, all files under the root directory are selected (thanks, @Gnossos #885, @cderv #886).
+
+## MAJOR CHANGES
+
+- Files with a leading `_` in their names are always ignored, even if they are specified in `rmd_files` in `_bookdown.yml`. Such files under subdirectories are also always ignored (#886).
 
 # CHANGES IN bookdown VERSION 0.18
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 - Text references do not work for `theorem` environments (thanks, @ssp3nc3r, rstudio/tufte#75).
 
+- when `rmd_subdir` and `rmd_files` are provided, only the `rmd_files` are now selected in addition to the `rmd_subdir` content. (thanks, @Gnossos, @cderv, #885).
+
 # CHANGES IN bookdown VERSION 0.18
 
 ## NEW FEATURES
--- a/R/utils.R
+++ b/R/utils.R
@@ -66,8 +66,10 @@ source_files = function(format = NULL, config = load_config(), all = FALSE) {
   subdir_files = setdiff(
     list.files(
       if (is.character(subdir)) subdir else '.', '[.]Rmd$', ignore.case = TRUE,
-      recursive = subdir_yes, full.names = is.character(subdir)),
-    files)
+      recursive = subdir_yes, full.names = is.character(subdir)
+    ),
+    files
+  )
   files = c(files, subdir_files)
   # keep only some wanted files
   if (length(files2 <- config[['rmd_files']]) > 0) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -62,13 +62,15 @@ source_files = function(format = NULL, config = load_config(), all = FALSE) {
   # a list of Rmd chapters
   subdir = config[['rmd_subdir']]; subdir_yes = isTRUE(subdir) || is.character(subdir)
   files = list.files('.', '[.]Rmd$', ignore.case = TRUE)
-  files = c(files, list.files(
-    if (is.character(subdir)) subdir else '.', '[.]Rmd$', ignore.case = TRUE,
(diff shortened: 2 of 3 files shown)


rickhelmus:
> I think we should always ignore files starting with `_`, otherwise the logic sounds too complicated to explain to users and also complicates the implementation. Thank you!

Hello,

This unfortunately breaks the recommendation of `pkgdown` to use underscores to cooperate with `bookdown` (see [https://pkgdown.r-lib.org/reference/build_articles.html](build_articles())). Is there still a way to include files with initial underscores?

Thanks,
Rick

yihui:
@rickhelmus I don't see how it breaks the recommendation of **pkgdown**. The help page says:

> `build_articles()` renders each R Markdown file underneath vignettes/ and saves it to articles/. There are two exceptions:
>
> - Files that start with `_` (e.g., `_index.Rmd`) are ignored, enabling the use of child documents in bookdown
>
> - Files in vignettes/tutorials are handled by build_tutorials()

So **pkgdown** also ignores files that start with `_`, which is consistent with **bookdown**. If you do want to use files that start with `_`, you can use them as [child documents](https://bookdown.org/yihui/rmarkdown-cookbook/child-document.html) of other files that do not start with `_`.

rickhelmus:
Hi yihui,

Thanks for your answer. I wasn't being really clear, sorry. I am actually already using one 'master' Rmd that includes several child documents. The problem is that I want to exclude this master document from `pkgdown`. Anyway, this may be more a feature request for `pkgdown` so I created a [bug there](https://github.com/r-lib/pkgdown/issues/1336).

Thanks,
Rick

yihui:
@rickhelmus Okay, now I see. Sorry about the trouble! Let me think more about it.

cderv:
Hi @rickhelmus , 

Just chiming in as I am trying to understand the interaction between pkgdown and bookdown in your usage.

> The problem is that I want to exclude this master document from pkgdown

As you are using pkgdown, I understand your working on a package. Then, I understand that you have a bookdown project inside your package that you store in the vignette folder but you don't want to get it rendered as article in your package doc website. Correct ?  
Or is is not a book project and are you just using one of the single document format (`bookdown::html_document2` ?) for you master.Rmd ? 

If you can provide a complete use case (maybe in a new issue) I am willing to dig into it to see what could be done on pkgdown or/and bookdown about this usage. 

However, @yihui seems to have a more clearer picture than mine though, so maybe this won't be necessary 😄

Issue #909, opened 2020-06-29
Links in ToC to sections within chapters fail when names are not unique


richarddmorey:
Creating two sections in different chapters with the same name appears to link all table of contents links back to the *first* section with that name. 

Steps to reproduce:

[bookdown-bug-demo.zip](https://github.com/rstudio/bookdown/files/4846370/bookdown-bug-demo.zip)

1. Download the demo here: https://bookdown.org/yihui/bookdown/get-started.html
2. In `01-intro.Rmd` add to the end:

```
## Visualisation

Test

```

3. In `02-literature.Rmd` add the same thing to the end (so the section labels are the same: "Visualization".
4. Build using the "Build" button.
5. Open the resulting HTML file and click on the table of contents link for the "Visualization" section of Chapter 3. Notice that it takes you to Chapter 2 instead.

What I expected: The link to the "Visualization" section in Chapter 3 should take you to Chapter 3.

Some things to note:
* Naming the "Visualization" section in `02-literature.Rmd` to "Visualization2" fixes the issue.
* I have confirmed this also occurs with the development version 0.20.1.
* Reverting to bookdown 0.19 fixes the issue.

[I note offhand that I found this issue when installing the development version to try to fix another issue last week. It slipped my mind reporting it because it turned out my other issue was with another package...]

```
> xfun::session_info('rmarkdown')
R version 4.0.0 (2020-04-24)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Catalina 10.15.5, RStudio 1.3.1023

Locale: en_GB.UTF-8 / en_GB.UTF-8 / en_GB.UTF-8 / C / en_GB.UTF-8 / en_GB.UTF-8

Package version:
  base64enc_0.1.3 digest_0.6.25   evaluate_0.14  
  glue_1.4.1      graphics_4.0.0  grDevices_4.0.0
  highr_0.8       htmltools_0.5.0 jsonlite_1.7.0 
  knitr_1.29      magrittr_1.5    markdown_1.1   
  methods_4.0.0   mime_0.9        rlang_0.4.6    
  rmarkdown_2.3   stats_4.0.0     stringi_1.4.6  
  stringr_1.4.0   tinytex_0.24    tools_4.0.0    
  utils_4.0.0     xfun_0.15       yaml_2.2.1     

Pandoc version: 2.9.2.1
```

---

By filing an issue to this repo, I promise that

- [X] I have fully read the issue guide at https://yihui.org/issue/.
[...]

PR #911, opened 2020-07-06
Enable pandoc `--file-scope` only when duplicate footnotes are detected

jjallaire:
By default, the `--file-scope` argument is now only passed to pandoc when duplicate footnotes are detected in the manuscript. This is controlled by the `bookdown.render.file_scope` option, which now defaults to `"auto"`. Set the option to `TRUE` or `FALSE` to force a particular behavior irrespective of whether duplicates are detected.

Addresses https://github.com/rstudio/bookdown/issues/909

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # CHANGES IN bookdown VERSION 0.21
 
-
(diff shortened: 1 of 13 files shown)


jjallaire:
Realized there are a couple of problems with this PR:

1) We had to introduce a `file_scope` option to the various formats because that's where the behavior is controlled in rmarkdown. However, it's not really an option (since anything the user specifies will be overridden), but more a sink where we can stick the results of our "auto" computation. 

2) The inclusion of `output_options` in the call to `rmarkdown::render()` will mask any `output_options` that would have been provided via `...`.

I think to do this properly we need to go back to `rmarkdown::render()` taking a `file_scope` argument. I'll work on that and close this PR in the meantime.

PR #912, opened 2020-07-06
pass file_scope to rmarkdown only when duplicate footnotes are detected

jjallaire:
By default, the `--file-scope` argument is now only passed to pandoc when duplicate footnotes are detected in the manuscript. This is controlled by the `bookdown.render.file_scope` option, which now defaults to `"auto".` Set the option to `TRUE` or `FALSE` to force a particular behavior irrespective of whether duplicates are detected.

Addresses https://github.com/rstudio/bookdown/issues/909

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.20.1
+Version: 0.20.2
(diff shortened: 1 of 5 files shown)


jjallaire:
Closing this as I think there are enough negative side effects of `--file-scope` that we should disable it entirely by default. Revised PR that does this to follow after I consult with @yihui on the best approach to this.

Issue #930, opened 2020-08-21
extra_dependencies and includes:in_header not working for PDF output.


jonathan-g:
Steps to reproduce:

1. Create a basic bookdown project using RStudio new project and choosing "Book project using bookdown"
2. Test that `bookdown::render_book("index.Rmd", bookdown::pdf_book())` builds correctly.
3. Edit `_output.yml` to add `extra_dependencies: "subfig"` to the `pdf_book` output:
```
bookdown::pdf_book:
  includes:
    in_header: preamble.tex
  latex_engine: xelatex
  citation_package: natbib
  keep_tex: yes
  extra_dependencies: "subfig"
```
4. Edit `04-application.Rmd` to add the example code from <https://bookdown.org/yihui/rmarkdown-cookbook/latex-subfigure.html>:
````
## Example one

```{r subfig-test, fig.cap='Figure 1', fig.subcap=c('(a)', '(b)', '(c)'), fig.ncol=2, fig.align="center", out.width="50%"}
plot(1:10)
plot(cars, pch = 19)
boxplot(Sepal.Width ~ Species, data = iris)
```

## Example two
````
5. Build using `bookdown::render_book("index.Rmd", bookdown::pdf_book())`.

## Expected result:

I expect a PDF document like the one in the example at https://bookdown.org/yihui/rmarkdown-cookbook/latex-subfigure.html

## Actual result:

XeLaTeX reports an error:
```
! Undefined control sequence
l.249 {\centering \subfloat
                           [(a)\label{fig:subfig-test-1}]{\includegraphics[w...

Error: LaTeX failed to compile Foo.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips. See Foo.log for more info.
```

After this error I tried again, adding the line 
```
\usepackage{subfig}
```
to `preamble.tex`, but I still got the same error.

Inspecting the generated `Foo.tex` file, there is neither an `include{preamble.tex}` nor a `\usepackage{subfig}`.

This error is similar to #888 

```
> xfun::session_info("bookdown")
R version 4.0.2 (2020-06-22)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363), RStudio 1.3.1056

Locale:
  LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
  LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
  LC_TIME=English_United States.1252    

Package version:
[...]

PR #931, opened 2020-08-23
Update documentation for render_book().

jonathan-g:
Add details about how the options are set for the output_format parameter to documentation for render_book() function. See discussion in comments to #930.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.20.2
+Version: 0.20.3
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
--- a/R/html.R
+++ b/R/html.R
@@ -673,6 +673,7 @@ parse_fig_labels = function(content, global = FALSE) {
 
   regmatches(content, m) = labs
 
+  write_ref_keys(names(arry))
   # remove labels in figure alt text (it will contain \ like (\#fig:label))
   content = gsub('"\\(\\\\#(fig:[-/[:alnum:]]+)\\)', '"', content)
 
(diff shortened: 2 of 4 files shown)


cderv:
Thanks ! 
What about adding examples as in the issue, in `\dontrun` clauses maybe ? one with the function & parameters and one with with the sring ? Could be best to prevent convention in the future, right ?

NEWS.md
```diff
@@ -7,6 +7,7 @@
 ## MINOR CHANGES
 
 - The `--file-scope` behavior introduced in bookdown v0.20 is now disabled by default. This is due to broken TOC links for duplicate section names (e.g., "Exercises"; see #909) that have automatically generated identifiers.
+- Updated documentation for `render_book` to make it clearer how options are set for the `output_format` parameter (See #930).
```
cderv, line 10: As below, syntax is 
```md
(thanks, @yourname, #prnumber)
```
Thanks!

jonathan-g:
I think @yihui makes sense in saying that the documentation fits better into rmarkdown, but it would be useful to put a modicum of documentation into bookdown alerting people to look at the rmarkdown documentation.

If you all want, I can write up some small PRs, one for rmarkdown documenting this behavior and the other for bookdown alerting the user to go look in rmarkdown for details.

I'll defer to what you all would like.

cderv:
Yeah it is right. This should be in `rmarkdown::render`. The help page of `render_book` already linked to `rmarkdown::render`, so having only this documented there seems ok to me. In `render_book` I would maybe only add some example with comments

```r
# will use the default format defined in _output.yml
bookdown::render_book("index.Rmd")
# will use the options for format defines in yaml
bookdown::render_book("index.Rmd",  "pdf_book")
 # If you pass an output format, it must have all the option set
bookdown::render_book("index.Rmd", bookdown::pdf_book(toc = FALSE))
```

and I would add the link to https://bookdown.org/yihui/bookdown/build-the-book.html in a reference part of this help page. 

We could also have `bookdown::render_book` inherits param for `rmarkdown::render` so that both are up to date, but it is not the practice here it seems - we just link to the other topic instead of inheriting.

jonathan-g:
Closed this and created a new PR  #933 following the suggestions from the discussion above. I also created a PR in rmarkdown (https://github.com/rstudio/rmarkdown/pull/1880)

PR #938, opened 2020-08-26
add number_sections argument to markdown_document2 (closes #756)

### LINKED ISSUE #756: Would you mind developing word_book() output format?
I was wondering if there would be any way that someone could develop `bookdown::word_book()` output format just like either `bookdown::pdf_book()` or `bookdown::html_book()`.
Currently, figure and table numbering does not follow chapter hierarchy when output type is `bookdown::word_document2()`, which means a figure inserted under Chapter 3 is numbered Figure 1 (not Figure 3.1) if that figure is the first one throughout all Rmds. This happens to table numbering as well.
It would be so much appreciated if `word_book()` format could be implemented.
Many thanks!

atusy:
Please merge after https://github.com/rstudio/rmarkdown/pull/1879

This PR adds number_sections argument to `markdown_document2` and its family (e.g., `word_document2`).

I set `TRUE` as a default value to be consistent with other formats (e.g., `html_document2`).
However, this may not be a good idea in case users' Rmd files manually numbered sections...

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.20.5
+Version: 0.20.6
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -61,7 +61,7 @@ License: GPL-3
 Imports:
     htmltools (>= 0.3.6),
     knitr (>= 1.22),
-    rmarkdown (>= 2.3),
+    rmarkdown (>= 2.3.5),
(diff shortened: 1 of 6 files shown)


yihui (approved):
@atusy Thanks!

@cderv Feel free to merge this one after rstudio/rmarkdown#1879 is merged.

cderv:
LGTM. 

Only one think: This feature relies on newer rmarkdown. So I did this 7f5af9c which is what to do in my experience. 
@yihui do you work that way or have you another process for this ? Waiting for you answer before merging. 

And we don't see this because ... we don't correctly test IMO. Once again, I think we should add test each time we add something. For now, I don't think the current test suite in bookdown allow to easily add test for this, so I'll do it later.

cderv:
It is not a perfect way to do it but I added a quick test because that is better to have one.

yihui (approved):
@cderv 7f5af9c looks good to me. Yes, we should use `Remotes` in `DESCRIPTION` when it requires a Github version of a package. Thanks!

atusy:
@cderv Thanks for changes!

cderv:
Thanks a lot @atusy !

PR #940, opened 2020-08-31
Custom environment using Fenced Divs

### LINKED ISSUE #924: Custom Environment Inline R code or nested chunks
Hi folks, so sorry to disturb again.  
1. Thanks so much for helping with [bookdown/issues/883](https://github.com/rstudio/bookdown/issues/883), which worked for me and resolved the issue.  
2. I posted on SO over a week ago, but seem to be getting very little traffic on my question. So i'd like to push it here, if you do not mind, please. [stackoverflow.com/bookdown-proof-custom-environment-how-to-nest-code-chunks-or-inline-r-inside](https://stackoverflow.com/questions/63116467/bookdown-proof-custom-environment-how-to-nest-code-chunks-or-inline-r-inside)
From my search, I believe and it looks like a lot of people have been looking for similar things over the years....

[As I posted there in the comments,](https://stackoverflow.com/questions/63116467/bookdown-proof-custom-environment-how-to-nest-code-chunks-or-inline-r-inside) I don't know if a child document or more likely a lua filter approach would work - despite using XieYiHui's awesome code for so many years now, I am still a total newbie unfortunately. Therefore, your expert help is all the more appreciated. [My only goal is to produce a book that can compile in pdf, epub, and html, BUT produce slides at the same time... in the original spirit of `beamer`...](https://stackoverflow.com/questions/52429949/creating-accompanying-slides-for-bookdown-project)

---

cderv:
This will resolve #924 

The aim is to use fenced Divs to create the custom block, instead of using special knitr's engines. Some notes in this PR

### Engines

* `eng_theorem` and `eng_proof` are in bookdown
* they are calling `knitr:::eng_block2`

### Tricky stuffs

* `\BeginKnitrBlock` are used but we may not need them anymore
* A special option is supported ("bookdown.output.markdown") to output as md... not sure we want to keep doing this.
* For latex, some stuff happens in `restore_block2()` - we should keep some
* Reference are dealt in bookdown base on a syntax `(#type:id)` - that means we need a unique id, even if not is provided. Currently chunk name is used but with fenced divs, pandoc can't access that. 
- [x] Check internationalization : https://bookdown.org/yihui/bookdown/internationalization.html ? 
    * For PDF it happens in `restore_block2()`
    * For HTML it happens in `parse_fig_labels` run in `resolve_refs_html`

### Conflicts

* See how this should work regarding `latex-divs.lua` in Rmarkdown. There is conflict.

### Current status

Something like this should work for now in this draft PR. It is not perfect but it is a start.

````md
---
title: "test2"
output: 
  bookdown::pdf_document2:
    keep_tex: true
    pandoc_args: !expr sprintf("--lua-filter=%s", here::here("inst/resources/lua/custom-environment.lua"))
  bookdown::html_document2:
    keep_md: true
    self_contained: false
    pandoc_args: !expr sprintf("--lua-filter=%s", here::here("inst/resources/lua/custom-environment.lua"))
---

```{theorem, pyth, name="Pythagorean theorem"}
For a right triangle, if $c$ denotes the length of the hypotenuse
and $a$ and $b$ denote the lengths of the other two sides, we have

$$a^2 + b^2 = c^2$$
```

It could be referenced: see \@ref(thm:pyth)

:::{.lemma #first-lem data-name="My first Lemma"}
Here is another, containing some R code
```{r}
1+1
```

and some markdown **bold text** with an inline `r "code"` and some equation on how
you calculate $x$ 

$$
x = y + z^2
$$ 

:::

Make sure that it could be reference \@ref(lem:first-lem)
````

There are a lot of things to check because of how things works currently: 

I'll keep working on this to make it work, and we can discuss the different choices here

## Must do  

- [x] Add tests ~- new testthat snapshoting may help here~ This will be a future project to use `expect_snapshot`
- [x] Make it option so that knitr engines still works - would be Pandoc Divs **or** knitr engines
- [x] Add a switch in rmarkdown do deactivate a lua filter or choose the execution order (solving the conflict between both filter)

### CODE REVIEW

**Comment by tchevri** [comment]
This is great, very exciting - I'll be happy to try it out whenever you think I won't be wasting your time, that's probably the most I can offer as clearly I don't play in your court when it comes to coding. This said, I don't know if it's relevant, but my bookdown creates also slides, as per a previous discussion. I noticed that when I have incremental bullets, it screws up the numbering! So on the first slide, a theorem may appear as theorem 3.1, but if the slide is repeated because of an incremental bullet, then this theorem incorrectly changes to 3.2 though it's the exact same theorem, not a desirable outcome imho. I thought i'd highlight it, as I am sure it will be an easy fix for you, but not something you might have necessarily spontaneously thought about.
Again, many thanks for your awesome progress, again, very exciting for all of us to see this feature finally happening.

**Comment by cderv** [comment]
Slides output are not really an obvious bookdown output, so yes I did not thought of that. I'll have a look at your example. 
But does it works correctly with the already existing theorem knitr engine ? Or is it more a feature request ?

Currently I am trying to reproduce what exists, but we can surely make improvement to make this more generic.

**Comment by tchevri** [comment]
True, but isn't it desirable to have some slides with your book? 
In particular if you want a textbook? 
Also, `beamer` / `beamerarticle` made the combination pretty smooth - so i naively thought this could be kept, but i was wrong, i think.
[However, XieYiHui helped me with creating slides as a bookdown output](https://stackoverflow.com/questions/52429949/creating-accompanying-slides-for-bookdown-project), it almost works perfect, just a few issues.
A bit of a hack of your topic here, I apologize, but as an fyi, issues remaining are
1. [Can't seem to be able to assign an output filename on the fly](https://stackoverflow.com/questions/52429949/creating-accompanying-slides-for-bookdown-project), this has been asked by others too, see comments.
2. [There was a nasty `pando`c "upgrade" that unexpectedly threw my code off and now forces me to stick with `pandoc 2.6`. I have received a few answers from John McFarlane, but nothing helpful that I know of unfortunately - it's been over a year.](https://stackoverflow.com/questions/57614387/pandoc-4317-forces-content-under-title-slides-to-be-included-in-a-frame-in-pand)
3. Custom environments, that's your current post.

So yes, right now everything works perfect for `knitr engine` - except for the ability to put code insides the ```` ```{} ``` ````, and that kind defeats the purpose, because the (my) goal is to have conditional content, hence the need for executable code to allow the switch (e.g. `` `r if (out_type=="beamer") "-"` ``). With regard to the current `knitr engine`, there is only this one issue that I ran into when inside the `theorem knitr engine`, I had incremental slides and saw that as the same slide was shown with incremental material, as I said, but the theorem numbering would also be incremented, which is clearly not desirable.

again - here is my opportunity to thank you very warmly, as i have very much benefited from all your work and help! Merci infiniment.

**Comment by cderv** [comment]
Thanks for the feedbacks. I'll take that into account. For new features, don't hesitate to open new issue if none already is. IMO It is easier to follow up on one feature per issue, than one topic issue with several bugs and feature request inside it.

**Comment by cderv** [comment]
So we get now rough support for `eng_proof` equivalent but it is not perfect

- [ ] Format is not exactly the same. Name should be before the dot, and the content should be right after the emphasize type text. This is a matter of `Blocks` vs `Inlines` I think 
- [ ] As `eng_proof` creates also a div, if it is used at the same time as the lua filter, there is a duplicated span line added. There should be a mechanism to either ignore this div block in some way, or do not activate the lua filter if those engines is used. 
- [ ] The id is removed from the _proof_ paragraph as not in the knitr engine. Maybe we could add it back - could be useful...
- [ ] Internationalization does not work yet. The mechanism for `eng_proof` and `eng_theorem` differs and the former does not currently play well with what we could do in the lua filter. 

Will keep 🤔 to fix these.

**Comment by cderv** [comment]
[#bookdown New custom blocks with Pandoc fenced div instead of specific bookdown engines](https://trello.com/c/MqXTEGJd/3-bookdown-new-custom-blocks-with-pandoc-fenced-div-instead-of-specific-bookdown-engines)

**Comment by cderv** [comment]
Two of the issues has been solved now. 

About the identifier, we add one for `eng_theorem` in order to reference but not with `eng_proof`. Maybe we can add and if to both but on the div instead of the span ? 🤔

**Comment by cderv** [comment]
Some more notes on the current issue

* _proof_ environment is really specific in Latex as provided by **amsthm** package. 
    * This means Providing a name to generate `\begin{proof}[name]` will lead to replacing _Proof_ by the provided name.  
    * There is no internationalization done in Latex currently for this type _proof_ (maybe a bug ?)

* Regarding internationalization
    * It works for PDF because it happens in the generated Tex preamble that is still adding by R
    * For HTML I see 2 solutions and a variant: 
        1. we can post process to detect `p><span id="my-proof" class="solution"><em>Solution</em> (A proof). </span>` and replace what is in the `<em>` tags by the translation provided in `_bookdown.yml`
        2. I think we can pass the `_bookdown.yml` as a metadata file to pandoc through `--metadata-file` command line option. That way we can get access to the value in the pandoc lua filter. 
        2. Variant: We could simply all the config file and get rid of `_bookdown.yml` and provide everything in the `index.Rmd` on a new `bookdown:` field. Those metadata would be accessible in the filter too. We could also use an external yml file to simplify and not clutter the index.Rmd file. 

I'll try b as it seems the cleaner and would open new possibility with lua filters.

**Comment by cderv** [comment]
For easier testing, I added the lua filter by default. 

One think I noticed: Currently we don't have good helper function in **rmarkdown** so that lua filters can be used in other package. I needed to add a function in bookdown, similar to `rmarkdown:::pandoc_lua_filters`. I think we should consider to improve this in **rmarkdown** so that it can be used in other 📦 

Anyway, we need now to decide when and where this will work: 

* Do we make this an opt-in mechanism ? Or an opt-out ? 
* Should it be with an option or with an argument passed to the format function ? 
* Do we want to support both `eng_proof` and `eng_theorem` with these new custom Divs ?
    * If so, then we need some more check. 
    * If not we may need a fail safe to detect both are in used and opt-out for example. 

@yihui If you have thoughts on this, do not hesitate to share. 

## Todos 

- [ ] Needs to figure out what to do with `latex-divs.lua` filter conflict. Maybe changing the order of the filter could help. 🤔 
- [ ] Add support for `markdown_document2` family too
- [ ] Tests more
- [ ] Document this new feature
- [ ] Add this in _bookdown\_demo_ (as it will help testing)

**Comment by cderv** [comment]
> Needs to figure out what to do with latex-divs.lua filter conflict. Maybe changing the order of the filter could help. 🤔

We really to make things better on **rmarkdown** side regarding where lua filter are called. Currently `pdf_document` and `html_document` does not insert the `--lua-filter` command line at the same place and hence there is no "not-hacky" way of making sure to call the bookdown lua filter first. 

I think we should target an improvment regarding the lua filter mechanism before considering merging this (as it is the first lua filter we introduce in bookdown)

**Comment by turtlegraphics** [comment]
Hi!  I'm currently writing a stat textbook with Bookdown.  We've got a mid-October deadline and realized a couple of weeks ago that fenced divs would solve all of our problems with exercises, solutions, and theorem style numbering.  I wrote a lua filter to replace latex-div.lua that implements fenced divs in both HTML and LaTeX.  It's probably not exactly what's needed for integration with the bookdown package, but we are using it now for a 400+ page book and it's working well.

It was a snap to handle the LaTeX side.  The HTML side was much harder, because the cross references needed to be done by hand. I've dealt with a lot of issues and pitfalls and maybe can be of service.

I've got a [test book](https://github.com/turtlegraphics/book-test) I'm using to build out technologies that our real book will use (like a single-chapter PDF makefile).  It's got the lua filter, fenced-blocks.lua, and you can see it in action by building the book.  You need to disable the built-in latex-div.lua, which I've done by simply commenting out its contents on my system.

I'm in the thick of the Fall semester and also trying to hand our book to CRC by mid-October, but I'm happy to help where I can.  You're welcome to use the code however you like.

**Comment by cderv** [comment]
Hi @turtlegraphics ! 

Thanks a lot for sharing your work ! This is really awesome ! There are a lot of great ideas ! 

I really like the numbering and cross referencing from the lua filter - we were not planning to do that from the lua filter but as you managed to do it, it could be interesting to add.  Also the special classes to allow further customisation in HTML is really cool !

We may integrate some of this and also fix some of your pitfall (latex-divs.lua conflicts, latex preamble or i18n). If I have question I'll ping you, but in the mean time it would be awesome if you find time to contribute here to make all this supported in bookdown (by testing or just by telling me which are the important feature to have so that I don't forget).

File: DESCRIPTION [hunk 1]
```diff
@@ -80,3 +81,4 @@ SystemRequirements: Pandoc (>= 1.17.2)
 LazyData: TRUE
 RoxygenNote: 7.1.1
 Encoding: UTF-8
+Remotes: rstudio/rmarkdown#1899
```
cderv, line 84: To remove after merging the PR in **rmarkdown**

cderv:
The conflicts with `latex-divs.lua` is now resolved as **bookdown** lua filter is added before the other and it deals with `data-latex` attribute if it sees it on one of the supported Divs classes, and removes it so that `latex-div.lua` in rmarkdown is not activated.

cderv:
We now got a function to translate from 
````md
```{theorem, name = "(something)"}
Some content
```
```{lemma, my-lemma, name = "(my Lemma)"}
Some content
```
````
to 
```md
::: {.theorem name="(something)"}
Some content
:::
::: {.lemma #my-lemma name="(my Lemma)"}
Some content
:::
```

But that is not that useful because any chunk using one of the theorem environment and that don't have a label and so won't have an id. That means, the fenced divs will currently be ignore. It is unexpected and makes it no so useful. 

This **must have an id** thing is really annoying !

cderv:
> This must have an id thing is really annoying ! But it is currently necessary for the labelling and internationalization for all eng_theorem related chunks.

Ok I now fixed that by generating an id for a div if none is provided, as `knitr:::unnamed_chunk()` does. 

It means we can write a fenced div with no id and it will render correctly in bookdown - `parse_fig_labels` will do its trick. Obviously, an id should be provided to be able to reference correctly using `\@ref`

Converting `02-components.Rmd` in `inst/examples` using `convert_to_fenced_div()` then rendering the bookdown book with version of this branch gives the correct result. The pandoc divs will just use `<p>` tags for the content, which we did not have before - so some spacing can appear differently. 

Using engine: 
image

After conversion, using fenced div:
image

This seems to impact the inserted blank square. So I need to look at that.

tchevri:
awesome - and interesting (if not weird) issue...
I believe most of us will be keen to know how you fix it, does not look so
obvious, why it even happens.
Thank you so much for all this effort - we all very much appreciate.

cderv:
it happens because of the css rule currently used. 
https://github.com/rstudio/bookdown/blob/57f1217c20835d538d103244ea46ec65acc374c3/inst/resources/gitbook/css/plugin-bookdown.css#L93-L96

As using pandoc, everything is inserted between a `<p>` tag, this rule creates a floating square that is not placed where we want. Modifying the rule to add to the last `<p>` child should work for the new syntax, but there won't be a square printed in html for the old syntax... Not sure it is bug loss...

cderv:
Ok this currently fails with pandoc 2.7.3 😞 - I was developing with 2.9.2.1 - I need to find why.
Testing on several pandoc versions like with Rmarkdown would be beneficial...

cderv:
Ok I tested several pandoc versions and I isolated the issue to be with pandoc 2.7.3 and earlier. It works from 2.8+

From there, following the [release note for 2.8](https://github.com/jgm/pandoc/releases/tag/2.8), I found that what is described in the documentation about Attributes element (https://pandoc.org/lua-filters.html#type-attr) only work for pandoc 2.8+

That means we need to explicitly use `pandoc.Attr` constructor and not the simpler syntax `{id, classes, attributes}`. 

It should work now with pandoc 2.7.3 - that was not easy to find as no explicit error, and it is not mention in the documentation.

cderv:
@tchevri I was looking back at your comment about beamer support, but I can't make the current engines work with beamer output. So I guess I won't deal with this here. could  you open a new issue specific to your beamer support idea with a working example so that I better understand  what is required. ok? Thank you !

cderv:
@yihui I think you can review now. 

* There is still the square placement for proof that won't work for previous book. 
* Where should we document this ? Bookdown book ? 
* We don't yet support epub, but I guess you can review what already works.

I tested on several pandoc versions, and with the bookdown book. 

Let's get your feedback before going any further - I already spent a lot of time on this one.

cderv:
Oh and one think I observed about the current behavior: When rendering to HTML, CSS for those environments like `div.theorem` are in `plugin-bookdown.css` used in `gitbook()` but not in `html_document2`. That means we don't get any styling for those environment by default. We could easily have a default. I'll open an issue maybe about that.

inst/resources/lua/custom-environment.lua
```diff
@@ -0,0 +1,197 @@
...
+    -- get the attributes
+    local options = div.attributes
+    if (options["data-latex"] ~= nil) then 
+        -- so that latex-divs.lua in rmarkdown does not activate
+        print("[WARNING] data-latex attribute can't be used with one of bookdown custom environment. It has been removed.")
+        options["data-latex"] = nil
+    end
+
+    -- return [name] for latex, and (name) for html
+    function get_name(format, options)
```
atusy, line 118: What about defining this function at the top level?
This function only uses local variables and given parameters.

```
local get_name = function(format, options)
...
```

Even if we need to define functions within a function, it is a good practice to define function locally like the above. Otherwise, `get_name` is defined globally and overwritten every time the `Div` function is called.

inst/resources/lua/custom-environment.lua
```diff
@@ -0,0 +1,197 @@
...
+        local name = options["name"]
+        if (name == nil) then return "" end
+        local template
+        if (format == 'latex') then
+            template = "[%s]"
+        elseif (format == 'html') then
+            template = " (%s)"
+        end
```
atusy, line 126: What about define

```lua
local template_list = {latex = "[%s]", html = " (%s)"}
```

at the top level, and then extract the value like

```lua
local template = template_list[format]
```

in the `get_name` function?

atusy (commented):
Good job! I briefly reviewed the lua filter, and suggest changes.

inst/resources/lua/custom-environment.lua
```diff
@@ -0,0 +1,197 @@
...
+    if not env_type.env then
+        print_debug("Not a bookdown supported custom class")
+        return div
+    end
+    print_debug("Found types -> ", env_type)
+
+    -- get the id if it exists - it will we use to build label for reference
+    local id = div.identifier
+    -- if no id, one is generated so that bookdown labelling mechanism works
+    if #id == 0 then id = unlabeled_div() end
```
atusy, line 104: can be simplified as

```lua
local id = div.identifier and unlabeled_div()
```

inst/resources/lua/custom-environment.lua
```diff
@@ -0,0 +1,197 @@
...
+    return "unlabeled-div-"..(counter)
+end
+
+-- Get metadata specific to bookdown for this filter
+Meta = function(m) 
+    bookdownmeta = m.bookdown
+    if (bookdownmeta ~= nil) then
+        -- For internationalization feature of bookdown
+        if (bookdownmeta.language ~= nil) then
+            if (bookdownmeta.language.label ~= nil) then
```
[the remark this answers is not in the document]
cderv: Great ! You are right, I discovered that [very recently ](https://github.com/rstudio/bookdown/blame/ce06eef1e09fa44abee4d5b475175678913e71f1/inst/resources/lua/custom-environment.lua#L95)and forgot to apply it here 😄 

I'll change that !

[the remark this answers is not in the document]
cderv: done in 088a0372202a9c0d0705a60b239e6545c6ba58c3

inst/resources/lua/custom-environment.lua
cderv: This is another way of doing it. The advantage is less line so easier to read.
But what would be the advantage to to define at the top level outside of `get_name` ? I would do that locally to the function only;

inst/resources/lua/custom-environment.lua
cderv: Does it work if `div.identifier` is `""` (empty string) and not `nil` ? 

Because I know I encounter this issue while working on that, and it may be here that I had an issue because otherwise I would have use `if (id ~= nil)` as in other places.

atusy: Oops, I'm wrong. Sorry.

```lua
> "" and 'foo'
foo
> "bar" and 'foo'
foo
```

inst/resources/lua/custom-environment.lua
atusy: The advantage is we can avoid running the identical code, which may improve performance slightly.
However, it should be subtle difference.
This is just a suggestion, not a request.

cderv: I removed the if clause, but keep the definition in the function as this variable is only interesting in this function
done in c45c9b867011f54012725119c74b0b55c74488e5

inst/resources/lua/custom-environment.lua
cderv: I moved it out and define it as a local function in c45c9b867011f54012725119c74b0b55c74488e5

atusy (approved):

inst/resources/lua/custom-environment.lua
cderv: Yeah that was the tricky thing and why sometime if was safer to test zero-lengh. But here it is because `div.identifier` does not return `nil` as I would have expected. 🤷

cderv:
Thanks a lot for the lua review @atusy !

atusy:
@cderv My pleasure!!

yihui (approved):
Sorry for the long delay! I just finished reviewing and tweaking this giant PR. I'll merge it for now, and take a look at the CSS issue later. Thanks!

inst/rmarkdown/lua/custom-environment.lua
```diff
@@ -0,0 +1,188 @@
...
+    
+    -- create the custom environment
+    local label
+    -- Create a label for referencing - only for theorem like env
+    if (env_type.type == "theorem") then
+        label = string.format("%s:%s", theorem_abbr[env_type.env], id)
+    end
+    print_debug("label for reference -> ", label)
+
+    -- TODO: should we support beamer also ?
```
yihui, line 133: Yes, I think we should.

inst/resources/gitbook/css/plugin-bookdown.css
```diff
@@ -90,7 +90,7 @@ div.theorem, div.lemma, div.corollary, div.proposition, div.conjecture {
 span.theorem, span.lemma, span.corollary, span.proposition, span.conjecture {
   font-style: normal;
 }
-div.proof:after {
+div.proof>p:last-child:after {
```
yihui, line 93: This has been addressed by c59e3dcc84058614bf30749ce8c8f26f2c97a6ee.

cderv: I kind of remember having tried that, and that was not ok in the output... I may have wrongly test it maybe. 🤷‍♂️

cderv:
Thanks! That was interested to look at the tweak you made to modify this PR. 

Even if you did not review the lua file per-se, I guess you've tested the filter itself and that worked ok for you ? Or you blindly trusted me with the general behavior of this new feature ? 😅 

> take a look at the CSS issue later. 

I opened an issue to track this: #977 but it seems you closed it already. 

And added another one for beamer support #978

yihui:
I only tested it briefly (in the bookdown-demo project).

cderv:
> A special option is supported ("bookdown.output.markdown") to output as md... not sure we want to keep doing this.

OK I finally found where `blogdown.output.md` is used: it is in **blogdown**
https://github.com/rstudio/blogdown/blob/master/inst/scripts/render_page.R#L10

We may want to check it is works with this new env once we support markdown output in the Lua filter.

yihui:
This option was introduced in e2142507eef26361643cae28d75dc0afdf06126b. The main purpose was to support theorem environments for the file format `.Rmarkdown` in **blogdown**. This is a rather minor use case. I doubt if anyone has written a single theorem in a `.Rmarkdown` post :)

That said, it will still be nice if the new syntax works for .Rmarkdown posts. Currently there is a missing step that I haven't done in **blogdown** yet: `.Rmarkdown` doesn't support bibliography. That will be implemented by converting `.markdown` to a certain Markdown output format. Theorems in fenced Divs could be converted to HTML in this step, too (`:::` needs to be converted to `<div>`).

PR #944, opened 2020-09-10
add a work_book alias to word_document2

cderv:
This will complement #756 

Help page in `?html_document2` precise
> These function are expected to work with a single R Markdown document instead of multiple documents of a book, so they are to be passed to rmarkdown::render() instead of bookdown::render_book(). 

But `word_document2` seems to work ok with a bookdown project with multiple files too. If I put `bookdown::word_book` in `_output.yml` it renders correctly. 

Does the new number sections addition change this limitation ? I can't remember why there was this limitation for `*_document2` format... 🤔


yihui (commented):
> These function are expected to work with a single R Markdown document instead of multiple documents of a book

The documentation was not accurate. I'll clarify the documentation in a moment.

Any of these `xxx2` formats actually do work with multiple Rmd files. The major difference with the `xxx_book` formats is that the latter usually generate multiple output files.

`word_book` generates one output file, so I'd rather not introduce this alias.

`pdf_book` exists because it serves as the basis of any PDF output formats (since it has a `base_format` argument). Perhaps `pdf_book` is not an intuitive name (it doesn't necessary generate books).

`epub_book` exists because EPUB is commonly known as a book format, although it doesn't have to be a book).

cderv:
> Any of these xxx2 formats actually do work with multiple Rmd files. The major difference with the xxx_book formats is that the latter usually generate multiple output files.

Makes sense but I did not get that from the documentation. 

> word_book generates one output file, so I'd rather not introduce this alias.

Like `pdf_book`... that is why I thought it would be a clarifying addition. But I see the intended usage of `pdf_book` with base format. 

Let's clarify the doc then and close this PR.

yihui:
Clarified via 74c8bef.

Issue #947, opened 2020-09-11
split by rmd puts some footnotes outside "page-inner" div


jtbayly:
I just changed a book to split by rmd instead of section, putting each section into its own file. The output for the bookdown::gitbook version takes each chapter's final section's footnotes and puts them outside the "page-inner" div, thus messing up the footnote styling for that section only. I've tried it on the default and dev versions of bookdown, as well as the packaged version of Pandoc and Pandoc 2.10.1. It's the same all around. 

Here's a simple demonstration book with the problem visible in sections 1.3 and 2.2, but all other sections working fine:
[bd-footnote-problem.zip](https://github.com/rstudio/bookdown/files/5210901/bd-footnote-problem.zip)

If you switch back to split by section and rebuild, the problem disappears.

xfun::session_info('bookdown')
```
R version 4.0.2 (2020-06-22)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6

Locale: en_US.UTF-8 / en_US.UTF-8 / en_US.UTF-8 / C / en_US.UTF-8 / en_US.UTF-8

Package version:
  base64enc_0.1.3 bookdown_0.20.6 digest_0.6.25   evaluate_0.14   glue_1.4.2     
  graphics_4.0.2  grDevices_4.0.2 highr_0.8       htmltools_0.5.0 jsonlite_1.7.1 
  knitr_1.29      magrittr_1.5    markdown_1.1    methods_4.0.2   mime_0.9       
  rlang_0.4.7     rmarkdown_2.3.5 stats_4.0.2     stringi_1.5.3   stringr_1.4.0  
  tinytex_0.25    tools_4.0.2     utils_4.0.2     xfun_0.17       yaml_2.2.1     
```

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.org/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('bookdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/bookdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
[...]

PR #964, opened 2020-10-04
Custom table of contents

cjdbarlow:
Allows a custom TOC file to be specified in _bookdown.yml, so component rmd files (and the order of) can be specified. This is mutually exclusive with other build options (e.g. rmd_subdir).

An example toc could be:
```
# Main
---
* [Home](index.Rmd)
* [Intro](content/01-intro.Rmd)
	* [Literature](content/02-literature.Rmd)
* [Methods](content/03-method.Rmd)
	* [Applications](content/04-application.Rmd)
		* [Summary](content/05-summary.Rmd)
* [Refs](content/06-references.Rmd)
```

- Warnings will be flagged to console if there are files in the toc that can't be found, or relevant files in the book directory that aren't included in the toc. It has been helpful typo detection.

Additionally, when used in the gitbook style:
- The "headers" (e.g. '#Main') are rendered on their own line in the TOC, and styled differently. This allows blocks of first-level TOC elements to be broken up visually into groups.
- Dividers ('---') can be inserted arbitrarily between TOC elements. Dividers break any indentation/sublists.
- Page names in TOC to be different from the actual page name (the H1 header). The [Name] is used for the TOC, whilst the H1 header on the page is still used on the page.
- Subsections are determined by indentation in the TOC (tab-separated), rather than by #/##/### etc. This means that #/##/### etc can be used for internal page styling, rather than determining TOC components.
- Best results seen/designed for/only tested when _output.yml has toc_depth: 1 and split_by: rmd, but this is not forced - I can see the toc being useful even if the gitbook style isn't used.

Other stuff:
- A 'stubs' folder is also specified, for holding bits of .rmd/.md files that are designed to be included in other files. This meets a criteria of my planned workflow. It affects how warnings are flagged (stubs are ignored), but no other material effects.
- A 'stringr' dependency has been 'added' - as rmarkdown already is dependent I think this is reasonable.


cderv:
Hi @chrisjake, Thanks for your contribution. 

In order to better understand the need of this, as there was no issue opened to discuss this prior to the PR, I am curious about why the classic header 1, header 2, ... structure that is usually reflecting in the TOC does not work for you ? 
Would it be because you want to have separate files for sub chapters ? 

IMO h1 / h2 / h3 represented by #/##/## are meant to structure the content first and not provide styling - there is css for that. 

I am just trying to better understand the problem you aim to solve with this PR before looking at it any further.

Thank you very much.

chrisjake:
Hi @cderv ,
There are two ways I see it being useful for my use case, which is migrating [this gitbook](https://partone.litfl.com/) to bookdown.

1. Separate files as subchapters (as you have said)
    - Allows subchapters to be moved around easily as the design changes (relevant for another project I have)
    - Increases resilience of the system to merge conflicts between multiple authors
    - Easier to manage long books with multiple subchapters
2. Customising the toc with headings (e.g. curriculum, in my example), and dividers

cderv:
> Customising the toc with headings (e.g. curriculum, in my example), and dividers

I think this is equivalent to the already supported special _(PART)_ header (https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html#special-headers) : Have you tried that already ? 

> Separate files as subchapters (as you have said)

Bookdown is currently design to have one file per header 1 chapter, otherwise it could lead to undesired issues with some features (like https://github.com/rstudio/bookdown/issues/947#issuecomment-692552387). Having the ability to be more free on how to organize the source file is an often asked question, and it is not just a matter of TOC. There is a splitting that happens from the rendered HTML to the multiple files - this is the challenge I think here.

But it is definitly a asked-feature we should look into.

chrisjake:
> I think this is equivalent to the already supported special _(PART)_ header (https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html#special-headers) : Have you tried that already ?

I had not seen that! That would provide the same function but also translate better to other formats; I will use that instead.

>There is a splitting that happens from the rendered HTML to the multiple files - this is the challenge I think here... definitly a asked-feature we should look into.

I agree; though I am not sure the best way to do this (YAML probably better than what I have done...) - I am not a software engineer by any stretch. Provided there is only 1 H1 header per file it seems to behave though, and I do like the flexibility it provides. I think functionality like this is valuable, even if this is not the mechanism by which it should be implemented.

carlaarchuleta707-lgtm (approved):

PR #966, opened 2020-10-05
Add an example for fontsettings config in the book

cderv:
As it is seems it was not obvious
https://community.rstudio.com/t/how-to-make-bookdown-dark-mode-by-default/83375/2


yihui (approved):
Thank you!

PR #969, opened 2020-10-05
Allow "url" to work with or without trailing /

jtbayly:
Adds the slash no matter what. If trailing slash is already there, the complete URI of the cover image will simply have an extra slash, which should be fine.

### DIFF
--- a/inst/templates/gitbook.html
+++ b/inst/templates/gitbook.html
@@ -11,7 +11,7 @@
   <meta property="og:title" content="$pagetitle$" />
   <meta property="og:type" content="book" />
   $if(url)$<meta property="og:url" content="$url$" />$endif$
-  $if(cover-image)$<meta property="og:image" content="$url$$cover-image$" />$endif$
+  $if(cover-image)$<meta property="og:image" content="$url$/$cover-image$" />$endif$
   $if(description)$<meta property="og:description" content="$description$" />$endif$
   $if(github-repo)$<meta name="github-repo" content="$github-repo$" />$endif$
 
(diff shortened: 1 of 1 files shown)


jtbayly:
I should also note that the following line appears to me to be wrong, but I left it alone:

```
            <i class="fa fa-circle-o-notch fa-spin"></i><a href="$if(book-url)$$book-url$$else$./$endif$">$if(title)$$title$$endif$</a>
```

There is no other place in the code-base that the variable "book-url" is used. I'm guessing this is supposed to be just "url". However, since I don't know what this is doing, I left it alone.

yihui (commented):
It's hard for me to make the decision. I agree that normally the extra `/` doesn't hurt, and users are far more likely to run into problems if the `/` is missing, so adding the slash seems to be a good idea. The only problem is, if the user really doesn't like the double slashes here, there wouldn't be a way to remove the extra one...

jtbayly:
They can actually remove the extra slash by removing the trailing slash from the end of the url in the yaml. Or is that not what you mean?

yihui:
I mean most users probably have had the trailing slash in `url`. With this PR, they will be forced to remove it. For this hassle and the hassle of adding the slash, I don't know which is worse.

If Pandoc's template has a little bit more programming capabilities, I'd test if `url` has a trailing slash, and add one if it doesn't. I guess that will solve all problems.

jtbayly:
I don't see why they will be forced to remove it. They won't notice it. It will silently continue to work. And it won't affect anywhere else that they might use the URL field, since it is only being added to these particular metatags.

I think the biggest difference in my mind, and the reason I lean toward including my patch (aside from personal pride, lol), is that it leaves books with silent (if non-optimal) success, whereas without it, perfectly valid URLs without a slash silently fail.

I leave it in your hands. Let me know which you go with. Either way there should probably be an update to the documentation, and I'll try to make a pull request for that once I know which way you go.

yihui (approved):
Okay. Let me just merge it for now, and see how it goes. Thanks!

yihui:
I just noticed that this PR made it impossible to specify a full URL as the cover image source, e.g.,

```yaml
cover-image: https://bookdown.org/yihui/bookdown/images/cover.jpg
```

I'll see how to fix it.

PR #981, opened 2020-10-14
Correctly relocate references with Pandoc 2.11

### LINKED ISSUE #976: Check references feature with new Pandoc --citeproc processing
From first tests based on the draft PR rstudio/rmarkdown#1916, the new citeproc processing does not output the same HTML as the filter did. This creates an issue with **bookdown** where references are not correctly relocated in each chapter as it should. 

I assume default style for citation is not the same. We may want to provide another default CSL if we don't want to change the default one.

_Originally posted by @cderv in https://github.com/rstudio/rmarkdown/issues/1915#issuecomment-707144003_

cderv:
This fixes #976 by modifying the regex to parse the references in HTML. 

Previously we had:
```html
<div id="refs" class="references">
<div id="ref-xie2015">
<p>Xie, Yihui. 2015. <em>Dynamic Documents with R and Knitr</em>. 2nd ed. Boca Raton, Florida: Chapman; Hall/CRC. <a href="http://yihui.name/knitr/">http://yihui.name/knitr/</a>.</p>
</div>
<div id="ref-R-bookdown">
<p>———. 2020. <em>Bookdown: Authoring Books and Technical Documents with R Markdown</em>. <a href="https://github.com/rstudio/bookdown">https://github.com/rstudio/bookdown</a>.</p>
</div>
</div>
```

 and now 
```html
<div id="refs" class="references csl-bib-body hanging-indent">
<div id="ref-xie2015" class="csl-entry">
Xie, Yihui. 2015. <em>Dynamic Documents with <span>R</span> and Knitr</em>. 2nd ed. Boca Raton, Florida: Chapman; Hall/CRC. <a href="http://yihui.name/knitr/">http://yihui.name/knitr/</a>.
</div>
<div id="ref-R-bookdown" class="csl-entry">
———. 2020. <em>Bookdown: Authoring Books and Technical Documents with r Markdown</em>. <a href="https://github.com/rstudio/bookdown">https://github.com/rstudio/bookdown</a>.
</div>
</div>
```

Some classes are added by Pandoc now when using `--citeproc`. 
I guess this allows for some customisation. 

This is note related to the issue, but let's note the span applied in the body have changed from 
```html
<span class="citation">(Xie <a href="#ref-R-bookdown" role="doc-biblioref">2020</a>)</span>
```
to 
```html
<span class="citation">(<a href="#ref-R-bookdown" role="doc-biblioref">Xie 2020</a>)</span>
```
So it applies by default to name and year, and not just year. 
I don't think this is an issue.

### DIFF
--- a/R/html.R
+++ b/R/html.R
@@ -348,7 +348,8 @@ split_chapters = function(output, build = build_chapter, number_sections, split_
   if (split_bib) {
     # parse and remove the references chapter
     res = parse_references(html_body)
-    refs = res$refs; html_body = res$html; ref_title = res$title
+    refs = res$refs; html_body = res$html
+    ref_title = res$title; refs_div = res$div
   }
   # parse and remove footnotes (will reassign them to relevant pages later)
   res = parse_footnotes(html_body)
@@ -415,7 +416,7 @@ split_chapters = function(output, build = build_chapter, number_sections, split_
     html = c(if (i == 1) html_title, html_body[i1:i2])
     a_targets = parse_a_targets(html)
     if (split_bib) {
-      html = relocate_references(html, refs, ref_title, a_targets)
+      html = relocate_references(html, refs, ref_title, a_targets, refs_div)
     }
     html = relocate_footnotes(html, fnts, a_targets)
     html = restore_links(html, html_body, idx, nms)
@@ -901,7 +902,8 @@ restore_appendix_html = function(x, remove = TRUE) {
 parse_references = function(x) {
   i = grep('^<div id="refs" class="references[^"]*">$', x)
   if (length(i) != 1) return(list(refs = character(), html = x))
-  r = '^<div id="(ref-[^"]+)".*>$'
+  regid = ' id="(ref-[^"]+)"'
+  r = sprintf('^<div%s.*>$', regid)
   k = grep(r, x)
   k = k[k > i]
   n = length(k)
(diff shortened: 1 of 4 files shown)


cderv:
In fact, I think we should relocate by creating a References div in each chapter but keeping the original classes (added by pandoc) so that if any CSS is applied, it also applies on the chapter. I am looking into that now.

cderv:
OK I have made the change now so that we get the same classes in the div for references in the chapter than the div at the end created by pandoc citeproc. This should allow to correctly style the divs in the document

R/html.R
```diff
@@ -901,7 +902,8 @@ restore_appendix_html = function(x, remove = TRUE) {
 parse_references = function(x) {
   i = grep('^<div id="refs" class="references[^"]*">$', x)
   if (length(i) != 1) return(list(refs = character(), html = x))
-  r = '^<div id="(ref-[^"]+)">$'
+  regid = ' id="(ref-[^"]+)"'
+  r = sprintf('^<div%s.*>$', regid)
```
yihui, line 906: How about this?

```suggestion
  r = '^(<div) id="(ref-[^"]+)"(.*>)$'
```

Then on line 912, use `  ids = gsub(r, '\\2', x[k])`

R/html.R
```diff
@@ -917,18 +919,20 @@ parse_references = function(x) {
   ref = paste(x[k], ref, x[k + 2], sep = '\n')  # add <div id=ref-...></div>
   title = if (grepl('^<h1[^>]*>', x[i - 2]) && grepl('^<div ', x[i - 3]))
     gsub('<span class="header-section-number">[.0-9]+</span>', '', x[i - 2])
-  x[k] = '<div>'  # remove the div id's
+  x[k] = gsub(regid, "", x[k])  # remove the div id's but keep the classes
```
yihui, line 922:
```suggestion
  x[k] = gsub(r, "\\1\\3", x[k])  # remove the div id's
```

yihui (approved):
I just suggested a minor change in the regex. Does that look okay? Thanks!

R/html.R
cderv: Oh I see. yes that looks of. Does the same but less lines 😄 

 Re reading it `[^>]*>` should be better than `.*>` also. 

I'll do the change.

cderv:
Done. My regex skill will improve! 😉

R/html.R
yihui: I think you are correct.

yihui (approved):
Looks perfect now!

Re: regex, believe it or not, I have used R for 16 years and read the help page `?gsub` for perhaps thousands of times, but I have never noticed `regexec()` until just a few days ago. It was something I wanted for several years, but I don't know why I missed this obvious function :)

cderv:
Oh I noticed `regexpr` `gregexpr` and `regexec` but I am still note sure how to use them. Something is not yet clicking 😄

yihui:
These three functions are often not directly useful by themselves, but can become super useful when combined with `regmatches()`.

cderv:
> regmatches()

I guess that is why that one too is not yet obvious to me... but with time it will 💪

PR #464, opened 2017-09-11
upgraded fontawesome font to v4.7.0

bernd-steinert-itc:
should fix #454

PR #984, opened 2020-10-15
Clarify that it is FontAwesome 4.7.0 in use

jtbayly:
The actual TTF file was updated to 4.7.0 a few years ago.


jtbayly:
I don't know if you guys want me to submit little stuff like this or not. I'm only doing it because I noticed that an included file claiming to be FontAwesome 4.1.0, but I remembered seeing a commit that updated it to 4.7.0. So I was confused and dug in to see which was actually in use. This pull request is the result of that investigation. It doesn't affect anything except possibly CDN caching, I guess.

cderv:
For reference, here is the PR where ttf file was updated to 4.7.0
https://github.com/rstudio/bookdown/pull/464

So yes I guess we should update the CSS file, should'nt we @yihui ?

yihui:
Sounds good! Thanks!

PR #990, opened 2020-10-21
Add default behavior for input in render_book by supporting a directory

### LINKED ISSUE #989: render_book() should handle more inputs
It should take a directory, so that `render_book("inst/examples")` can work. If you supply a directory instead of `index.Rmd`, `render_book()` could automatically look for `index.Rmd` and error if it can't find it.

Then the `input` argument could default to `.` so that `bookdown::render_book()` would just work when you're in a book project.

cderv:
This will close #989 

Currently 

* `input` has no default value, but when it is not provided (i.e `render_book()` only), it will error. 
* `render_book` will assume it is run from the book project directory which is the working dir. One must use 
```r
xfun::in_dir("inst/examples", bookdown::render_book("index.Rmd"))
```
from the bookdown project for example 

This PR adds the ability to provide a directory to `render_book` (as is `serve_book`) and in that case it will use `index.Rmd` in that directory if it exists. 

* New default for `input` is current working directory, so `index.Rmd` will be used if it exists. 
* If a directory is provided, working is changed for that directory for the rendering (like in `serve_book`)
* There was not much assertation on that `input` argument so I added files' existences tests to stop early in case any of them does not exists. 

Now, with this PR, this is possible
```r
bookdown::render_book("inst/examples")
# or
xfun::in_dir("inst/examples", bookdown::render_book())
```

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -47,12 +47,19 @@
 #' bookdown::render_book("index.Rmd", bookdown::pdf_book(toc = FALSE))
 #' }
 render_book = function(
-  input, output_format = NULL, ..., clean = TRUE, envir = parent.frame(),
+  input = NULL, output_format = NULL, ..., clean = TRUE, envir = parent.frame(),
   clean_envir = !interactive(), output_dir = NULL, new_session = NA,
   preview = FALSE, config_file = '_bookdown.yml'
 ) {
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
Looks good. Thanks!

PR #993, opened 2020-10-21
remove improper og:url set in templates

### LINKED ISSUE #970: meta property="og:url" set improperly
The meta property "og:url" should be set to the canonical URL of the *page* you are currently on, not the canonical url of the *site.* (https://ahrefs.com/blog/open-graph-meta-tags/)

Setting the og:url to something other than the current page [tells FB](https://stackoverflow.com/a/7831012/8802249) (and others?) to completely ignore all the data on the current page and scrape the og:url instead.

People should be able to share the particular page of the book they are reading. This doesn't completely prevent that. The link is still correct, but it does prevent the correct *information* from being displayed on FB, forcing every link to a particular page in the book shared on FB to display the info from the /index.html page instead.

You can see the details of what happens if I share the following link to the [YAML Options page](https://bookdown.org/yihui/bookdown/yaml-options.html) of the bookdown book on FB on FB's sharing debugger [here](https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fbookdown.org%2Fyihui%2Fbookdown%2Fyaml-options.html).

---

jtbayly:
fixes #970 by simply not setting og:url.


yihui (approved):
Re: https://github.com/rstudio/bookdown/issues/1308#issuecomment-1067097383, this PR looks okay to me (I don't remember why it fell off the radar). Two things to check:

1. Whether this works with Twitter: https://cards-dev.twitter.com/validator
2. Whether this could impact our scraping script for bookdown.org: https://github.com/rstudio/bookdown.org/blob/main/R/create-md.R

@cderv Could you verify? Thanks!

cderv:
Also, I am still unsure that `og:url` should be removed. What I am seeing is this: 
From https://ogp.me/
> og:url - The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "https://www.imdb.com/title/tt0117500/".

Is the canonical url of a HTML page in an online book the book url or the specific content page ? The object for me is the HTML book but maybe each html page should be considered as its own object. 

Facebook documentation gets more precise
https://developers.facebook.com/docs/sharing/webmasters?locale=en_US
> The [canonical URL](https://developers.facebook.com/docs/sharing/webmasters/getting-started/versioned-link) for your page. This should be the undecorated URL, without session variables, user identifying parameters, or counters. Likes and Shares for this URL will aggregate at this URL. For example, mobile domain URLs should point to the desktop version of the URL as the canonical URL to aggregate Likes and Shares across different versions of the page.

About canonical url 
https://developers.facebook.com/docs/sharing/webmasters/getting-started/versioned-link
> All links that specify the same canonical URL are treated as the same resource by Facebook's link sharing infrastructure for the purpose of metrics and metadata (link image, title, etc.). You can view engagement metrics for a URL using the [Graph API](https://developers.facebook.com/docs/graph-api/reference/url) call.

So for me, canonical url should be the book for which you want to aggregate the data. 

But maybe I am completely mistaken and this should indeed be removed. I don't want to remove the option if others feel like it should be set the current way. 🤔 

This is my current thinking on this.

jtbayly:
I understand the concern about changing the current behavior, but honestly, the aggregation of "likes" for a book is meaningless compared to actually displaying the correct information when you are sharing a chapter of a book. 

The whole point of having separate HTML pages for separate chapters is so that you can do things like share a link to something other than an entire book. It's the same as sharing a link to an individual blog post instead of the homepage of the site the post was on. 

Is the benefit of showing "15 shares" somewhere instead of "5 shares" really worth the confusion of showing the wrong info when you share it? If I wanted to share a link to the whole book, I'd share a link to the index.html page, not a chapter, or subsection of a chapter. 

Furthermore, the examples you cite from the FB documentation come after the first and most basic part of the documentation:

> an HTML meta tag to specify the canonical URL **for the page**. [emphasis mine, obviously]

I don't see any way to argue that the canonical URL for one of the HTML pages is actually a completely different HTML page. The fact is that the canonical URL for chapter 6 is the link to chapter 6, not the link to chapter 1 or the TOC. 

None of the examples of times when you'd want to point to a different URL apply. We're not disambiguating separate URLs to the same content. We're inexplicably pointing to *different* content and claiming it's the *real* place to find the shared content. 

There *may* be times when og:url should be different from meta canonical, but I really don't think this is one of those times. 

In summary, I'm very opinionated. ;)

And there's no reason to display the wrong information about the link you're sharing. It's not going to take them to the book, (or the homepage of a blog.) It's going to take them to the specific chapter of the book, (or the specific post of the blog) that you wanted to share. That's the info we should show. 

As always, of course, I understand the decision is yours in the end. Also, I've been wrong before. :)

Oh, and I'm sorry, but I don't have a current demo site running the modified version.

jtbayly:
One other note. I do agree that to some degree, there's a sense in which a book is more of a single entity. But the proposed solution actually accounts for that. It still shows the title of the book, along with the chapter info. The only thing it doesn't do is aggregate shares. And the number of shares for the book isn't even possible for us to show on our book anyway. I'm guessing it's entirely a theoretical benefit.

jtbayly:
Yeah, I can't say "nobody is making use of the aggregated counts." I just doubt it. I'm not by any means a FB expert, but here are my guesses. 

To make use of it *outside* of FB, you'd have to be calling the FB API to find out what the number of shares for the book is. I've not seen anybody using the FB API in or about a bookdown book. Maybe somebody has, but again, I kind of doubt it.

To make use of it *inside* FB, I'm not sure you could do anything. I don't think FB generally shows the number of shares for a link, but rather the number of shares of a post. It's been a while, and I could be wrong about this. In terms of benefit beyond that, I think you'd have to assume that the greater number of shares to the "same" URL would lead to FB prioritizing it for more displays or something. Whereas I expect it could just as easily be the opposite: That the misleading nature of hiding the information for the page they'll actually land on leads FB to deprioritize the link. 

Regarding the YAML, yes, we still need access to the URL for the entire book. Not sure if you're asking something else?

yihui (approved):
This sounds like a tough decision to make. I'm going to merge this PR for now, and can definitely revert it in the future if anyone complains about it (we can also provide an option for users to decide whether to include this `<meta>` tag). In the meantime, users can always add this line back with the `includes: in_header` option, e.g., 

```yaml
bookdown::gitbook:
  includes:
    in_header: meta.html
```

where `meta.html` is

```html
<meta property="og:url" content="..." />
```

Thank you both!

cderv:
Thanks @yihui ! 

And thanks @jtbayly for the discussion and the PR !

jtbayly:
Great. Hopefully I’m right and it won’t be a problem.

PR #994, opened 2020-10-21
bring default.html in line with gitbook.html template

jtbayly:
I think #969 should have included this change as well, to make both templates work the same way with cover images.


yihui (approved):
That's okay. Thanks!

PR #1008, opened 2020-11-04
rename bootstraplib package to bslib

cpsievert:
(To be merged after https://github.com/rstudio/bootstraplib/pull/150)


hadley:
Merged this to avoid breaking code in the wild

Issue #1016, opened 2020-11-09
bs4_book: add a font size for h4?


maelle:
Currently in the CSS for bs4_book there are font sizes for h1, h2, h3 but not h4 so if one uses an h4 it is bigger than h3.

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.org/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('bookdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/bookdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

hadley:
That'd be great. Want to do a PR?

maelle:
:heavy_check_mark: https://github.com/rstudio/bookdown/pull/1019

PR #1018, opened 2020-11-09
bs4_book: make it possible to override CSS?

maelle:
Related to #1016 as I was trying to add a quick tweak to my own book.

Or if all tweaks are supposed to be made via the theme argument, maybe the css argument of bs4_book (in the ... ) should be somehow coupled with warning message? (I realize forbidden might be too much as I use the css argument myself for print styles :grin:)

Cc @hadley

### DIFF
--- a/inst/templates/bs4_book.html
+++ b/inst/templates/bs4_book.html
@@ -10,11 +10,6 @@
   <meta name="author" content="$author-meta$" />
   $endfor$
 
-  <!-- CSS -->
-  $for(css)$
-  <link rel="stylesheet" href="$css$" />
-  $endfor$
-
   <!-- JS -->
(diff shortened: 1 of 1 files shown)


hadley:
How are you supplying this argument?

maelle:
It was in the css argument, I had written it in the `_output.yml` config.

I had

```yml
bookdown::bs4_book:
  theme:
    primary: "#1f58a3"
    yellow: "#804600"
  includes:
    before_body: WIP.html
  css: style.css
```

Now I have

```yml
bookdown::bs4_book:
  theme:
    primary: "#1f58a3"
    yellow: "#804600"
    h4-font-size: "1.1rem"
  includes:
    before_body: WIP.html
```

which seems more correct of course given my tweak could be made with a bootstrap variable. However, when using the CSS file, I was expecting it'd be loaded after the other stylesheets.

hadley:
Ah makes sense. As you work on this, any documentation improvements to `bs4_book()` would also be appreciated 😄 We should probably add a "Customising" section to the topic.

maelle:
Just a note that I see some overlap (in concepts) with https://rstudio.github.io/distill/website.html#theming since it also uses variables.

PR #1029, opened 2020-11-20
Add GHA workflow

cderv:
This PR adds GHA setup 

One for R CMD CHECK on the R package, on 3 OS, with 3 R versions on linux and using several pandoc versions, current RStudio 1.3 version (2.7.3), version in RStudio 1.4 (2.11.2 for now), and last devel version. 

One for building the book which is in this repository. I tried to understand what is you current workflow, but I am really not sure. 
* Is the book deploy to github pages still useful ? The book lives on bookdown.org now, and the deployment happens with R 
https://github.com/rstudio/bookdown/blob/be72ac4c7a963a3254da6a05db04e5608ef72a27/inst/examples/_render.R#L43
* Is the deployment on bookdown.org manual ? I believe you need credential for `rsconnect` and it seems you have them locally only
https://github.com/rstudio/bookdown/blob/be72ac4c7a963a3254da6a05db04e5608ef72a27/.gitignore#L4
* There is a trick in YAML to have all formats build when `[build all]` is in the commit message. 
https://github.com/rstudio/bookdown/blob/be72ac4c7a963a3254da6a05db04e5608ef72a27/.travis.yml#L20
Is this still useful to you ? 

I wanted to make a clean workflow for all this, but I am not sure of what is still important, and what is your workflow with all this. So, currently it will just build the book in the `gitbook` format. This would be a test to see that the build of the book is not breaking.

I believe we need to discuss this before merging.


cderv:
Testing on newer pandoc help us see differences. Here with Pandoc 2.11.2 atx headers are now the default - hence why the tests are failing.

cderv:
@yihui waiting for you to have a look about the publishing process of the book

yihui (approved):
Looks good. Thanks!

Replies to the three questions below:

> * Is the book deploy to github pages still useful ?

No.

> * Is the deployment on bookdown.org manual ?

Yes, and it will be great if there is a way to automatically publish to bookdown.org from GHA (the key is figure out the authentication with RStudio Connect).

With 0b313c7d389c2c2c341b0d7d03fe3938cc3ef5d0, I don't think I need to commit the `rsconnect` folder.

> * There is a trick in YAML to have all formats build when `[build all]` is in the commit message

That trick is no longer useful, but as I said in the previous comment, I wish we could build all formats on GHA and publish to bookdown.org automatically.

PR #1035, opened 2020-12-01
add 404 page for bs4_book and Gitbook

### LINKED ISSUE #1025: bs4_book: add a default 404 page?
If there's no 404.html file after building the book, would it make sense to add one that has the book TOC and some standard message?

@hadley 

---

maelle:
Fix #1025 

* A bit hacky
* Maybe the message on the page should be customizable


maelle:
preview for an example

image

cderv (commented):
I think we can do something for gitbook also and close #1046 too. 
I'll take it from here but work on your branch in this PR

cderv:
So I went further with my idea with this. 

* 404.html page is now added in the HTML splitting process. This mean it works for gitbook() and bs4_book()
* If a `404.html` exist in the main repo at the root level with the Rmd files, then it will be used. I don't think this should be used as none of the header, footer and sidebar will be included in this case. This is mainly there to avoid overriding an existing file if it is there.
* If a `_404.Rmd` or a `_404.md` is there, it will be render as a `html_fragment` and the content will be included as the body of the 404 page. This means all the theme (navbar, footer, sidebars, ...) will be there.
* If none of the above is there, then a default 404.html is created using a simple content (a header, and a body of 2 paragraphs)
* Currently, there is no way to opt-in or out of this behavior. 

@apreshill before going further with docs and tests, is this something as you imagined ? 
We could definitely adapt and simplify.

Demo : 

* gitbook: https://cderv.github.io/bookdown-demo/error => https://github.com/cderv/bookdown-demo/blob/master/_404.md
* bs4_book: https://cderv.github.io/bs4booktesting/error => https://github.com/cderv/bs4booktesting/blob/master/_404.Rmd
* Default content: https://github.com/rstudio/bookdown/pull/1035/files#diff-cfc07c2b017dee6a6f4b1a97fc55474bab2f97d8ca247779a1cfdd0929809538R466

cderv (changes requested):
Some changes to made following live review with @apreshill 

Overall we are good with this feature

R/bs4_book.R
```diff
@@ -346,6 +346,12 @@ bs4_chapters_tweak <- function(output,
     message("Tweaking ", path)
     index[[i]] <- bs4_chapter_tweak(path, toc, rmd_index = rmd_index, repo = repo)
   }
+  # tweak 404.html ---
+  path_404 <- file.path(output_dir, "404.html")
+  if (file.exists(path_404)) {
+    message("Tweaking ", path_404)
+    bs4_chapter_tweak(path_404, toc, rmd_index = rmd_index, repo = repo)
```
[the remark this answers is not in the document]
cderv: @apreshill  

Done in 93894e7

=> https://cderv.github.io/bs4booktesting/error

R/html.R
```diff
@@ -440,6 +444,38 @@ split_chapters = function(output, build = build_chapter, number_sections, split_
   nms[j]
 }
 
+build_404_page <- function(html_head, html_toc, html_foot, build, ...) {
+  path_404 = "404.html"
+  if (file.exists(path_404)) return(path_404)
```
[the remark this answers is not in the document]
cderv: From 92c5a72

We do nothing if `404.html` exist and assumes user has already something in place to move the file into the output directory. File will be untouched.

R/bs4_book.R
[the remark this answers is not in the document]
apreshill: Looks great! ⭐

cderv (approved):
Only thing left is documentation in the book.

cderv:
@yihui Just a friendly ping so that it is on your radar to have a quick thoughts and review from you before deciding to merge this. We can discuss it live tomorrow if you have time to have a look. Thanks! 

Behavior of this PR is described here: https://github.com/rstudio/bookdown/pull/1035#issuecomment-872289224

R/html.R
```diff
@@ -440,6 +444,40 @@ split_chapters = function(output, build = build_chapter, number_sections, split_
   nms[j]
 }
 
+build_404_page <- function(html_head, html_toc, html_foot, build, ...) {
+  path_404 = "404.html"
+  # if a 404 page already exist, we do nothing specific and assume
+  # user has already a workflow in place
+  if (file.exists(path_404)) return()
```
yihui: If the 404 page exists, we should also move it via `move_to_output_dir()` above, right? (If you `return()`, it won't be moved)

R/html.R
```diff
@@ -440,6 +444,40 @@ split_chapters = function(output, build = build_chapter, number_sections, split_
...
+    xfun::Rscript_call(function() {
+      rmarkdown::render(found,
+                        rmarkdown::html_fragment(
+                          pandoc_args = c("--metadata","title=404")
+                        ),
+                        output_file = path_404,
+                        quiet = TRUE
+      )
+    })
+    html_404 = Filter(nzchar, xfun::read_utf8(path_404)) # remove empty line
```
yihui: What's the goal of removing empty lines?

yihui (approved):
@cderv Just two small questions. Thanks!

R/html.R
cderv: If there is already a 404.html, then it should not be moved I think but it could be copied to output_dir. 
If a 404.html exists, then it should be consider a source file and not moved to output dir (which could be cleaned after that with `clean_book()`, or could be gitignore). Moving would delete right ? 

However, it could be copied. We assumed that if someone has already a 404.html in a bookdown project then there is already a custom way in the project to make this move to the output dir. But it would indeed be more interesting to have this internalize in bookdown. Do you think it is safe to do this ? 

Currently we found one example where a 404 page exists and we did not want to override any mechanism they could have from moving the file. It is for https://github.com/jtr13/EDAV/ but from their source and from https://edav.info/publish.html#make-a-custom-404-page, it is highly possible that the 404 page is copied to output dir just because they linked it in their doc and it is seen as a resource. 

So we could definitely offer to automatically copy to output_dir a 404.html existing in the source folder. 

Does it make sense ?

R/html.R
cderv: Some empty lines are introduced at the top of the `404.html` created with `html_fragment()`. I believe this is because of the template for `html_fragment()` which has empty lines and if conditional. 

So I wanted to remove them are they are not needed.

R/html.R
```diff
@@ -452,12 +452,10 @@ build_404_page = function(html_head, html_toc, html_foot, build, ...) {
...
-    xfun::Rscript_call(function() {
-      rmarkdown::render(
-        found, rmarkdown::html_fragment(pandoc_args = c('--metadata', 'title=404')),
-        output_file = p404, quiet = TRUE
-      )
-    })
+    xfun::Rscript_call(rmarkdown::render, list(
+      found, rmarkdown::html_fragment(pandoc_args = c('--metadata', 'title=404')),
+      output_file = p404, quiet = TRUE
+    ))
```
cderv: When I tested it worked as expected. No error regarding `p404` (or `path_404` before that) being not found. That is why I kept it this way. 

but I cannot make it work with a small example so not sure how it worked. 

It is safer to do this - I'll remind that.

cderv (commented):
Thanks for you changes ! 

That makes a lot of sense to tweak the function this way. Also `existing_files()` is a useful function ! 😅 Hard to know them all 😉 

I guess you found the idea interesting. That left us with 
* the choice of what should we do with an existing `404.html`: should we copy it (https://github.com/rstudio/bookdown/pull/1035#discussion_r672950361) in output dir ? 
* We need to document this. Where in the book ? Or in which function's help page ? I can do it once we decide.

R/html.R
yihui: Okay, let's just leave it untouched and assume that if 404.html exists, authors know how to copy it.

R/html.R
yihui: In this case, these empty lines don't really matter, right?

Sometimes empty lines do matter, e.g., inside `<pre></pre>`. I tend not to apply `Filter(nzchar)` here. What do you think?

yihui:
For documentation, perhaps we could add a section in Chapter 6: https://bookdown.org/yihui/bookdown/publishing.html This topic seems to be too small to fit a whole section, though.

cderv:
> For documentation, perhaps we could add a section in Chapter 6

This could indeed be a good place to put it in the Github publishing section as 404.html is supported by it. We don't have a Netlify part yet, but it will also be supported. However, no matter where you publish a 404 page will be created though. 

I was thinking maybe into: 
* Chapter 4: https://bookdown.org/yihui/bookdown/customization.html because there will be a default and you can customize it using `_404.Rmd`. But it is not cross format. 
* Chapter 3.1: https://bookdown.org/yihui/bookdown/html.html would be a good place too because this is a feature for HTML books only, which will apply on all HTML book format. It could be documented once in Gitbook, and mentioned again in a new bs4_book section with a link back to the Gitbook subsection ? Or a new "3.1.4 Common element" but it seems to short for that.

R/html.R
cderv: I see. Empty lines seems to not be an issue when in source code - The browser will remove them when they are between HTML tags. I'll remove this then.

cderv: Done in a8d98d9

yihui (approved):
We can discuss the documentation later. This PR looks good to me now.

cderv:
I opened a new issue to track the documentation. 

And I'll merge this one then. 

Thank you !

PR #1036, opened 2020-12-01
Greater flexibility in repo specification

### LINKED ISSUE #1012: Greater flexibility in repo specification
Also optionally need to specify subdirectory and branch name (important for master vs main). I think making `repo` a list would make this easiest to specify in yaml:

```
bookdown::bs4_book:
  repo: 
    base: https://github.com/hadley/ggplot2-book
    branch: master
    subdir: book
```

cc @maelle

maelle:
Fix #1012


kulbhushanchand:
Patiently waiting for the merger of this pull-request.

maelle:
Maybe I should add tests first. :thinking:

R/bs4_book.R
```diff
@@ -615,3 +664,43 @@ bs4_check_dots <- function(...) {
     )
   }
 }
+
+# test helpers -----------------------------------------------------------------
```
maelle, line 668: it's plural but there's only one for now :-)

R/bs4_book.R
```diff
@@ -390,16 +415,35 @@ tweak_tables <- function(html) {
...
     repo_edit <- NULL
     repo_view <- NULL
   }
 
-  template_link(html, ".//a[@id='book-repo']", repo)
+  if (!is.null(repo$base) && !grepl("github\\.com", repo$base)) {
+    template_link_icon(html, ".//a[@id='book-repo']", "fab fa-gitlab")
+    template_link_icon(html, ".//a[@id='book-source']", "fab fa-gitlab")
+    template_link_icon(html, ".//a[@id='book-edit']", "fab fa-gitlab")
+  }
```
cderv, line 444: Why don't we offer a parameter to choose the fa icon to use ? 
instead of just assuming non github = gitlab ?

R/bs4_book.R
cderv: is this only for test ? Meaning test helpers are stored inside the source of package ? 
We usually store theme inside the test folder directly. 

Also this function duplicates with `bookdown::book_skeleton()`, or the internal `bookdown_skeleton()` used for template. Maybe we can adapt those for the need of `bs4_book()`. It should be just a matter of rendering with special format (`render_book(output_format = bookdown::bs4_book())`)

cderv (commented):
I just add a quick look but here are a few comments. 

Thanks!

R/bs4_book.R
maelle: So the function should actually

* call `bookdown::book_skeleton`
* modify files if needed (content, YAML configs)
* call `render_book`?

cderv: > modify files if needed (content, YAML configs)

Do you need to have special content in the Rmd for your test ? You should be able to control most of the config behavior from `render_book()`

tests/testthat/helper-bs4_book.R
```diff
@@ -0,0 +1,24 @@
+
+create_minimal_bs_book <- function(path,
```
maelle, line 2: The current recommendation is to store helpers in R/ but this works too (I personally prefer it) https://blog.r-hub.io/2020/11/18/testthat-utility-belt/#code-called-in-your-tests

tests/testthat/helper-bs4_book.R
```diff
@@ -0,0 +1,24 @@
...
+                                   name = "book",
+                                   title = "Awesome Cookbook",
+                                   author = "Yoda",
+                                   metadata = NULL,
+                                   output_options = NULL) {
+  book_skeleton(
+    name = name,
+    title = title,
+    author = author,
+    index_metadata = metadata,
```
maelle, line 12: in the end the current test doesn't use it but I feel it might be useful?

cderv (changes requested):
That looks great. I made a few comments. 

About tests for this special format that works only on suggested package, do you know if CRAN expect tests to work only with Imports package or Suggests is ok ? 

If they don't I believe we may need to add checks in tests or skip on CRAN all the testthat part so that is complies. 

About xml2 used in testing for HTML output, is this something you have done in pkgdown also ? 

Also, do we need to add something into the bs4_book() doc about `github-repo` field in index.Rmd YAML ? Do we need to set this metadata value to the same value as in this repo field ?

R/skeleton.R
```diff
@@ -56,19 +58,25 @@ book_skeleton = function(
...
-    write_file(content, rmd_files[i])
+    write_file(content, file.path(path, rmd_files[i]))
   }
   write_file(
-    sprintf('bookdown::%s: default', c('gitbook', 'pdf_book', 'epub_book')), '_output.yml'
+    sprintf('bookdown::%s: default', c('gitbook', 'pdf_book', 'epub_book')),
+    file.path(path, '_output.yml')
   )
-  write_file(sprintf('book_filename: %s', name), '_bookdown.yml')
+  write_file(sprintf('book_filename: %s', name), file.path(path, '_bookdown.yml'))
```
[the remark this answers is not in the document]
maelle: why change the directory rather than using `file.path()`?

maelle:
> That looks great. I made a few comments.

Thanks a lot for the feedback!

> About tests for this special format that works only on suggested package, do you know if CRAN expect tests to work only with Imports package or Suggests is ok ?
> If they don't I believe we may need to add checks in tests or skip on CRAN all the testthat part so that is complies.

Good point, I have added a helper skipper. Cf also https://github.com/r-lib/testthat/issues/1398

> About xml2 used in testing for HTML output, is this something you have done in pkgdown also ?

Definitely. [Example 1](https://github.com/r-lib/pkgdown/blob/master/tests/testthat/test-autolink_html.R), [example 2](https://github.com/r-lib/pkgdown/blob/master/tests/testthat/test-html-tweak.R).

> Also, do we need to add something into the bs4_book() doc about github-repo field in index.Rmd YAML ? Do we need to set this metadata value to the same value as in this repo field ?

But `bs4_book()` does not use `github-repo` at all at the moment? Even in #1034 I haven't added. I actually do not know what this meta property does, actually. Maybe it should be tackled in that other PR?

R/skeleton.R
```diff
@@ -56,19 +58,25 @@ book_skeleton = function(
...
+      default_metadata <- list(
+        title = title,
+        author = author,
+        documentclass = documentclass
+      )
+      index_metadata = modifyList(default_metadata, as.list(index_metadata))
       content = c(
-        '---', sprintf('title: "%s"', title), sprintf('author: "%s"', author),
-        sprintf('documentclass: "%s"', documentclass),
-        'site: bookdown::bookdown_site', '---', '', content,
```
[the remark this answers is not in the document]
maelle: Ouch, sorry and thank you!

R/skeleton.R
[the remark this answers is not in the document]
cderv: I find it just help avoid repetition (every writing needs to `file.path(path, ...)`, it must not be forgotten) and DNRY is something good., Also it could make it easier to read 

* `xfun::in_dir("my_dir",  something_in_another_dir())`
* `withr::with_dir("my_dir", something_in_another_dir())`

But that is all. No benefit in the R side of things. It work the same. 

Let's leave it like that

cderv (approved):
That is ok. I'll take it from here.

cderv:
@yihui just so you know: 

As discussed this includes new testthat infrastructure for `bs4_boo()` testing. Everything is skipped when either deps are not installed or Pandoc is not available. So tests should be run on CRAN if both condition are TRUE. This is as  we dicussed I believe but we can skip on CRAN if it causes trouble somehow.

For the rest, it concerns the `bs4_book()` output only, excep a small tweak in `bookdown::book_skeleton()` to support a different path than working dir.

yihui:
Sounds good. Thanks for letting me know!

PR #1073, opened 2021-02-02
make serve_book() support the new bs4_book format

homerhanumat:


cderv:
Thanks for catching that!

Issue #1090, opened 2021-03-04
Using multiple filepaths in the `input` argument with `output_format = "all" errors


shirdekel:
I'd like to use both multiple file paths in the `input` argument to `render_book()` (in order to supply multiple dependencies for `drake::knitr_in()` in my `drake` plan), as well as specify `output_format = "all"`. However, just doing the former gives me warnings that suggest that only the first file path is being used, and supplying both gives me an error. I struggled making a minimal reprex, so instead cloned the [bookdown-demo repo](https://github.com/rstudio/bookdown-demo) and executed the following:

# Works  

``` r
bookdown::render_book('index.Rmd')
```

``` r
bookdown::render_book('index.Rmd', output_format = "all")
```

# Produces warnings

``` r
bookdown::render_book(c('index.Rmd', '01-intro.Rmd'))
```

``` r
Warning messages:
1: In if ((n1 <- nchar(p)) == 0) return(x) :
  the condition has length > 1 and only the first element will be used
2: In if (is_sub_path(p, d, n2)) { :
  the condition has length > 1 and only the first element will be used
3: In if (p2 == "") p2 = "." :
  the condition has length > 1 and only the first element will be used
```

# Produces error

``` r
bookdown::render_book(c('index.Rmd', '01-intro.Rmd'), output_format = "all")
```

``` r
Error in base::file(file, encoding = "UTF-8") : 
  invalid 'description' argument
```

# Session info
``` r
R version 4.0.2 (2020-06-22)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6

Locale: en_AU.UTF-8 / en_AU.UTF-8 / en_AU.UTF-8 / C / en_AU.UTF-8 / en_AU.UTF-8

Package version:
  base64enc_0.1.3   bookdown_0.21.6   digest_0.6.27     evaluate_0.14    
  glue_1.4.2        graphics_4.0.2    grDevices_4.0.2   highr_0.8        
  htmltools_0.5.1.1 jsonlite_1.7.2    knitr_1.31        magrittr_2.0.1   
  markdown_1.1      methods_4.0.2     mime_0.10         rlang_0.4.10     
  rmarkdown_2.7.2   stats_4.0.2       stringi_1.5.3     stringr_1.4.0    
  tinytex_0.29      tools_4.0.2       utils_4.0.2       xfun_0.21        
  yaml_2.2.1       
```

---

By filing an issue to this repo, I promise that

[...]

PR #1091, opened 2021-03-04
Add a vectorized wrapper for xfun::relative_path

cderv:
To handle several inputs in `render_book(input=)`. Currently it seems to work but we have warnings thrown in `xfun::relative_path`
```r
Warning messages:
1: In if ((n1 <- nchar(p)) == 0) return(x) :
  the condition has length > 1 and only the first element will be used
2: In if (is_sub_path(p, d, n2)) { :
  the condition has length > 1 and only the first element will be used
3: In if (p2 == "") p2 = "." :
  the condition has length > 1 and only the first element will be used
```
Reported in #1090 

Other solution would be to make `xfun` functions vectorized. @yihui would that be better ? 

This type of situation make me feels we are trying to recreate / maintain / tests some functionality we could get elsewhere : **fs** package is rather low level with no dependency and could help power all the path handling in the packages. Just throwing the idea in case this is a project we want to tackle.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,8 @@
 
 - Added the same CSS as in default Pandoc's template for when a CSL is used (#1045).   
 
+- No more warnings are thrown when passing several input files to `render_book(preview = TRUE)` (#1091).
+
 ## MINOR CHANGES
 
 - `anchor_sections = TRUE` becomes the default for `bookdown::gitbook()`.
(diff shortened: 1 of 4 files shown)


yihui (approved):
> Other solution would be to make `xfun` functions vectorized.

Yes. I'll merge this PR for now and think about it later.

> This type of situation make me feels we are trying to recreate / maintain / tests some functionality we could get elsewhere

It's always a trade-off between introducing new dependencies and maintaining one's own code. If the latter is not too complicated, I tend to go the latter way. **fs** does contain a huge set of functions about file paths, but I only need a very small subset of it.

Thanks!

yihui:
FYI `xfun::relative_path()` can take a vector now (**xfun** v0.22 was released to CRAN last night).

PR #1096, opened 2021-03-10
Move css to inline HTML style. CSS ineffective without #header.

adamvi:
By moving the CSS inline we keep the style changes without the CSS.  This can be changed with CSS override by users - see [here](https://css-tricks.com/override-inline-styles-with-css/).

The proposed changes result in the following output:

###  Single author, no affiliation (most basic - note it will change the spacing between author and date from the original)

Single_auth_no_affil

###  Multiple authors, no affiliation (color background added for context)

Auth_no_affil_proposed

###  Multiple authors with affiliation (color background added for context)

Auth_w_affil_proposed

### DIFF
--- a/inst/templates/gitbook.html
+++ b/inst/templates/gitbook.html
@@ -123,7 +123,7 @@ $if(subtitle)$
 $endif$
 $for(author)$
 $if(author.name)$
-<p class="author" style="margin: 1em 0 -0.5em 0; background: yellow;"><em>$author.name$</em></p>
+<p class="author" style="margin: 1em 0 -0.5em 0;"><em>$author.name$</em></p>
 $if(author.affiliation)$
 <address class="author_afil">
 $author.affiliation$<br>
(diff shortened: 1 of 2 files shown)


inst/templates/gitbook.html
```diff
@@ -145,7 +137,7 @@ <h2 class="subtitle"><em>$subtitle$</em></h2>
 $endif$
 $endfor$
 $if(date)$
-<p class="date"><em>$date$</em></p>
+<p class="date" style="margin-top: 1.5em"><em>$date$</em></p>
```
yihui, line 140:
```suggestion
<p class="date" style="margin-top: 1.5em;"><em>$date$</em></p>
```

yihui (approved):
Great. Thanks!

yihui:
@RLesur Sorry for the trouble! Please feel free to comment whenever you find any of our decision is inappropriate. I'm considering reverting this change now (#1170). Thanks!

Issue #88, opened 2016-04-15
brief flicker when navigating to new pages while using dark theme


kevinushey:
Any chance the theme-change flicker could be avoided when navigating across sections?

![flicker](https://cloud.githubusercontent.com/assets/1976582/14552851/abd5fd8c-0296-11e6-9bb7-1853371fb117.gif)

yihui:
The theme is applied after the page is loaded, and I think that is why it was white initially for a short moment. I tried to apply the theme as early as possible via JavaScript, but I was unable to completely avoid the flicker. The fundamental solution is probably write the CSS class in HTML at bookdown render time, but that is not really a solution since I don't know which theme the reader is going to use.

kevinushey:
One other minor UI issue to add for this thread -- when expanding + collapsing the side bar while a dark theme is in use, a white bar is shown between the empty space during the transition. Can a global background color be set on the `<body>` element to match the current theme, or something similar?

![whitespace](https://cloud.githubusercontent.com/assets/1976582/21661129/70868c4e-d287-11e6-9e8d-ac70d5ecfb11.gif)

yihui:
That might be an easier problem to solve than the original one, but I probably won't be able to work on it until after rstudio::conf.

cderv:
I had a look in this issue and indeed the transition background was not so tough to fix. 

For the brief flicker, it is a different issue. Would setting CSS variable to change using JS be better ? 
Not sure because it would still be JS to load at the right time.

PR #1100, opened 2021-03-15
Set background color for transition in one of the color mode

cderv:
This avoid a white background in sepia or night mode and is fixing one of the issue mentioned in https://github.com/rstudio/bookdown/issues/88#issuecomment-270501556


yihui (approved):
Looks good. Thanks!

PR #1103, opened 2021-03-18
toc_bib works with biblatex

### LINKED ISSUE #450: pdf_book(toc_bib=T) does not work when using biblatex
When set 
```
output:
  bookdown::pdf_book:
    citation_package: biblatex
```
the `toc_bib` para does not work.

I guess this is duo to that biblatex uses `\printbibliography` to print bibliography and thus cannot be matched by `r = '^\\\\bibliography\\{.+\\}$'` (`line 189` from `R/latex.R`).

When using biblatex, adding bib to toc could be achieved by manually editing the line `\printbibliography` in the `tex` file to `\printbibliography[heading=bibintoc]`.

cderv:
This will close #450 and also fixed `toc_bib` that was no more working because of indentation in tex template. 

Basically, natbib and biblatex does not include the same syntax. See Pandoc's template: https://github.com/jgm/pandoc/blob/master/data/templates/default.latex
which was explained in the issue.

Also, the trick for natbib does not work for bibtex. We need to use another syntax. 
I choose to modify the tex line with regex to add the option `heading=bibintoc`.
Other solution would have been to add `\DeclarePrintbibliographyDefaults{heading=bibintoc}` in preamble. This seems more complicated from within `add_toc_bib()` function and it works only with recent biblatex (https://tex.stackexchange.com/a/544718/209358)

I added test that help me fix already some edge cases.

I tested manually on bookdown-demo using

* `citation_package: biblatex` or `citation_package: natbib`
* with `biblio-title: References` in index.Rmd

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 ## BUG FIXES
 
+- In `pdf_book()`, `toc_bib = TRUE` now works with _natbib_ and _biblatex_ as `citation_package` (thanks, @qifei9, @umarcor, #450).
+
 - CSS dependencies like `url('truetype/Spectral-ExtraLight.ttf')` (with single quotes) are now correctly identified and moved (thanks, @RLesur, #991).
 
 - `fenced_theorems()` now correctly transforms implicit label in chunk header to a fenced divs id. (thanks, @enixam, #982)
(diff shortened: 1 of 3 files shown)


yihui (approved):
I didn't review the code very carefully, and I'm just going to trust you :) Thanks!

Issue #952, opened 2020-09-22
Chapter-bib issues with footnote citations and CSL files


jimhshen:
I opened an unanswered [question on stackoverflow](https://stackoverflow.com/questions/63852602/bookdown-how-to-make-citations-in-footnotes-appear-in-the-chapter-references) a while ago about citations in footnotes missing from end of chapter bibliographies.

A workaround to footnote citations (discovered by Lars) is to include an invisible chunk containing the citations such that they are rendered in the chapter bibliography, [seen here](https://jimhshen.github.io/citationfootnotes/workaround-footnote-citations.html).

```
# Workaround Footnote Citations

The workaround is to include an invisible text chunk.

This is a citation [@johndoe2020].[^note2]

This sentence has another footnote citation.^[This is another footnote that contains a citation @bobjohnson2018]

[^note2]: This footnote contains a citation [@janesmith2019]

::: {.invisible style="display:none"}
@bobjohnson2018
@janesmith2019
:::
```

Another issue I have discovered with chapter-bib citations is that it seems to have issues with CSL files (specifically the chicago-author-date.csl file found on the [Zotero styles repository](https://www.zotero.org/styles)).

A minimal example can be seen [here](https://jimhshen.github.io/citationfootnotes/chapter-bib-issues.html). The [repository is here](https://github.com/jimhshen/citationfootnotes). The YAML header for index.Rmd (where the CSL is defined) is below.

```
--- 
title: "A Minimal Book Example"
author: "Yihui Xie"
date: "`r Sys.Date()`"
site: bookdown::bookdown_site
output: bookdown::gitbook
documentclass: book
bibliography: [book.bib, packages.bib, test.bib]
csl: chicago-author-date.csl
link-citations: yes
github-repo: rstudio/bookdown-demo
description: "This is a minimal example of using the bookdown package to write a book. The output format for this example is bookdown::gitbook."
---
```

The following is the bib file

```
@article{bobjohnson2018,
	title = {Article 3},
	author = {{Bob Johnson}},
	year = {2018}
}


@article{janesmith2019,
	title = {Article 2},
	author = {{Jane Smith}},
	year = {2019}
}

@article{johndoe2019,
	title = {John's first paper},
	author = {{John Doe}},
	year = {2019}
}

@article{johndoe2020,
	title = {John's second paper},
	author = {{John Doe}},
	year = {2020}
}


@article{able2001,
	title = {An entry by multiple authors},
	publisher = {whatever},
[...]

PR #1111, opened 2021-03-30
Also parse references in footnotes

cderv:
This PR solves #952 by modifying the order in which looking for the targets ids. 

1. We first gets href= in <a> targets to retrieve relevant footnote
2. We then rerun the function to get href in body + relocated footnote
3. We use the full identified ids to find the references and append to body
4. we rerun footnote relocation so that footnotes are appended after references

Tested with the reproducible example given: https://github.com/jimhshen/citationfootnotes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 ## BUG FIXES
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
Thanks!

Issue #1114, opened 2021-03-30
[FR] internationalization - Customize figure names to set label after number 


tamas-ferenci:
(Already posted in Stackoverflow [here](https://stackoverflow.com/questions/54158401/customize-figure-name-just-as-chapter-name-in-bookdown) to no avail.)

Bookdown [allows the customization][1] of chapter names, which is handy when a Hungarian project is built, as `Chapter 1` is translated as `1. fejezet` (i.e. the renaming of `Chapter` is not enough).

However, the very same is true for figure (and table) names as well, because `Figure 1.1` should be translated as `1.1. ábra`, but I don't know how to achieve this (`_bookdown.yml` doesn't seem to support something like "`language/ui/figure_name`" just as it does with `chapter_name`).

There is no problem with the PDF (LaTeX) output, because including `\usepackage[magyar]{babel}` in the `preamble.tex` solves this problem, but I don't know what to do with the non-PDF formats, where my only idea was to use the `_bookdown.yml`.

  [1]: https://bookdown.org/yihui/bookdown/internationalization.html

tamas-ferenci:
Because the translation of Figure should go _after_ the number, not before it (as I have written: "`Figure 1.1` should be translated as `1.1. ábra`") and this is not possible with the possibilities listed in that doc, where I can only change the text, but I can't put it after the number in my understanding.

cderv:
That is right, this is currently not possible to do that. Thanks for explaining. 

It will always be `<fig_lab>: <number>` for now. 

Out of curiosity, in which language do you label figure this way ?

tamas-ferenci:
It's Hungarian. The reason why I hope for a possible new feature to support this is that the exact same thing is already working for chapter name! The cited doc even gives an example for this. The very same solution would be perfect for figure (and table) names.

cderv:
You mean the use of a function taking chapter number as argument ? Is that what you use for chapter name ?

tamas-ferenci:
Yes, `Chapter 1` is translated as `1. fejezet`. The solution is therefore simply:

```
language:
  ui:
    chapter_name: !expr function(i) paste0(i, ". fejezet")
```

A similar solution would be perfect for figure and table naming.

yihui:
That sounds like a reasonable feature request to me.

tamas-ferenci:
Thanks @yihui ! Together with https://github.com/rstudio/bookdown/issues/671 it'd make bookdown very nicely working for Hungarian as well :)

[...]

PR #1120, opened 2021-03-31
Allow use of a function for fig, eq or tab labels

cderv:
This will solves #1114 

For now I only added support for `fig`, `eq` and `tab`. Support for theorem or proof env is a bit trickier because `label_prefix` is used for find the label to use for `\newtheorem` in `preamble` where number is not used
https://github.com/rstudio/bookdown/blob/cec8efb9cc695d4c8e873f828abe9e171bd5d50e/R/latex.R#L239
and for proof env too
https://github.com/rstudio/bookdown/blob/cec8efb9cc695d4c8e873f828abe9e171bd5d50e/R/latex.R#L248

This means it can't be a function of reference number, and case is different between LaTeX and non LaTeX. 
I wanted to support all because for non LaTeX output this is the same mechanism: label is prepended by default. However, I did not find an easy way to do it because of the constraint mention above. But also I am not sure that what I have done currently is a lot easier. 

@yihui you may want to review that more closely in case you see a way to also allow the use of a function for all `theorem_abbr` and `label_names_math2`.

### DIFF
--- a/R/ebook.R
+++ b/R/ebook.R
@@ -117,8 +117,10 @@ resolve_refs_md = function(content, ref_table, to_md = output_md()) {
         if (type %in% theorem_abbr) {
           id = sprintf('<span id="%s"></span>', j)
           sep = ''
+          label = paste0(id, label_prefix(type), ref_table[j], sep, ' ')
+        } else {
+          label = label_fun(type, sep = sep)(ref_table[j])
         }
-        label = label_fun(type, sep = sep)(ref_table[j])
         content[i] = sub(
           m, paste0(id, label, ' '), content[i]
         )
--- a/R/ebook.R
+++ b/R/ebook.R
@@ -117,10 +117,8 @@ resolve_refs_md = function(content, ref_table, to_md = output_md()) {
         if (type %in% theorem_abbr) {
           id = sprintf('<span id="%s"></span>', j)
           sep = ''
-          label = paste0(id, label_prefix(type), ref_table[j], sep, ' ')
-        } else {
-          label = label_fun(type, sep = sep)(ref_table[j])
         }
+        label = label_prefix(type, sep = sep)(ref_table[j])
         content[i] = sub(
           m, paste0(id, label, ' '), content[i]
         )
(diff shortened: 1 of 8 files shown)


R/ebook.R
```diff
@@ -114,13 +114,13 @@ resolve_refs_md = function(content, ref_table, to_md = output_md()) {
       if (grepl(m, content[i])) {
         id = ''; sep = ':'
         type = gsub('^([^:]+).*$', '\\1', j)
-        if (type %in% names(theorem_abbr)) {
+        if (type %in% theorem_abbr) {
```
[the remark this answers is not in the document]
yihui: Sounds good. Thanks for catching it!

R/latex.R
```diff
@@ -234,18 +234,24 @@ restore_block2 = function(x, global = FALSE) {
   # fenced div (\begin) is used
   if (length(grep(sprintf('^\\\\(BeginKnitrBlock|begin)\\{(%s)\\}', paste(all_math_env, collapse = '|')), x)) &&
       length(grep('^\\s*\\\\newtheorem\\{theorem\\}', head(x, i))) == 0) {
+    theorem_label = vapply(
+      lapply(theorem_abbr, label_prefix),
+      function(fun) fun(), character(1), USE.NAMES = FALSE)
```
yihui, line 239: Is it equivalent to this?

```suggestion
    theorem_label = vapply(theorem_abbr, function(a) {
      label_prefix(a)()
    }, character(1), USE.NAMES = FALSE)
```

yihui (approved):
I think it is much more common for users to customize figure and table names. We can consider theorems in the future if someone asks for the feature. Thanks!

R/latex.R
cderv: Oh yes! Thanks - don't know how I missed that. Sometimes I do too complicated 😅

tamas-ferenci:
Thank you! Unfortunately theorem is the same story in Hungarian (`Theorem 1` should be `1. tétel`), but it is of course needed much-much less often than figure or table, and I am personally fine now. Thank you again!

cderv:
Yes theorem was trickier to support. 
Do you change it already for pdf using LaTeX in preamble ?

PR #1122, opened 2021-04-01
Allow the knit button to work in subdirs

### LINKED ISSUE #1121: preview_chapter() does not work with Rmd files in a sub-directory
preview_chapter works as intended for previewing Rmd files that are in the root directory. However, it does not render the same Rmd file that is within a subdirectory.

I'm not sure whether the inability to preview Rmds in subdirectories is intentional or not. For our purposes, we are putting Rmd content in a subdirectory within the root folder, so that the root folder doesn't become overwhelmed. It would be nice to be able to preview chapters without needing to render the whole book.

Here is a reprex using the [bookdown-demo](https://github.com/rstudio/bookdown-demo) as an example; note that I've used `render_book(preview = TRUE)` rather than `preview_chapter` but should be functionally equivalent. Hopefully it makes sense, I can clarify anything if needed.

```r
# Using https://github.com/rstudio/bookdown-demo as an example
# While all content in the root directory
bookdown::render_book("index.Rmd", "bookdown::gitbook") # works
bookdown::clean_book(TRUE)
bookdown::render_book("01-intro.Rmd", "bookdown::gitbook", preview = TRUE) # works
bookdown::clean_book(TRUE)

# Now move the book contents to a subdir = content
# Need to add rmd_subdir: ["content"] and rmd_files: ["index.Rmd"] to _bookdown.yml
bookdown::render_book("index.Rmd", "bookdown::gitbook") # works
bookdown::clean_book(TRUE)
bookdown::render_book("content/01-intro.Rmd", "bookdown::gitbook", preview = TRUE) # blank html pages
```

cderv:
fixes #1121 

It is like with **blogdown**, with this adjustment a Rmd file in a subdir can be previewed using the knit button. 

Only issue is that RStudio IDE won't open the preview because 

* `rmarkdown::render()` returns a relative path e.g `_book/02-litterature.html` to the bookdown project
* RStudio IDE consider a relative path as being relative to the input file - which in this case is not because the Rmd file is in a sub directory. 

Rmarkdown website deals with that by suppressing message for `render()` to avoid `Output created:` to be written by `render` in console and throw itself the message with `render_site()` - this happens in `default_site()`. 

I tried adding one message in `render=` inside `bookdown_site()` but as `render()` already emits one, only the first is taken into account it seems.

We could do the same in **bookdown** but it seems a bit too much to suppress message from `render()` completly. 
What about a new option to introduce in **rmarkdown** that would allow us to deactivate the printing of this message in `render()` from other package ? 

@yihui what do you think we could do ? 

I put as draft so that we decide what to do, but it seems making the Knit button work for a bookdown project is important. I recognize that using subdirs and preview may not be a lot of cases.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 - In `_bookdown.yaml`, `label` fields `fig`, `tab` and `eq` can now take a function as `ui` fields `chapter_name` and `appendix`. This function takes the reference number as only argument and must return a character to be used as full label. The default is a string prepended before the reference number. This new feature gives more flexibility to change the default for other language, e.g append the label name after the number. See updated doc about [Internationalization](https://bookdown.org/yihui/bookdown/internationalization.html)(thanks, Tamás Ferenc, #1114)
 
+- Using the 'Knit' button now also works with a Rmd file in a sub-directory of the book project (when `rmd_subdir` is used in `_bookdown.yml`) (#1122)
+
 ## BUG FIXES
 
 - Adapt CSS in `gitbook()` and `html_book()` for correct diplaying of `<details><summary>` content (thanks, @maelle, #971)
--- a/R/site.R
+++ b/R/site.R
@@ -10,8 +10,9 @@ bookdown_site = function(input, ...) {
 
   on.exit(opts$restore(), add = TRUE)
 
-  # need to switch to the input directory for computing the config
-  oldwd = setwd(input)
+  # need to switch to the project directory for computing the config
+  book_proj = find_book_proj(input)
+  oldwd = setwd(book_proj)
   on.exit(setwd(oldwd), add = TRUE)
 
   # load the config for the input directory
@@ -33,7 +34,10 @@ bookdown_site = function(input, ...) {
     if (is.null(input_file)) {
       in_dir(input, render_book_script(output_format, envir, quiet))
     } else {
-      render_book(input_file, output_format, envir = envir, preview = TRUE)
+      res = xfun::in_dir(book_proj, {
+        render_book(input_file, output_format, envir = envir, preview = TRUE)
+      })
+      res
     }
   }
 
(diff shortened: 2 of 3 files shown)


yihui (approved):
> What about a new option to introduce in **rmarkdown** that would allow us to deactivate the printing of this message in `render()` from other package ?

I have wished to have a way to suppress only the "Output created" message in the past, too, but I hesitated to add one more argument to `rmarkdown::render()` just for this purpose. I just thought about it again, and perhaps we could use `withCallingHandlers()` to capture the message, suppress it, and signal it later. Here is an example:

```r
writeLines(c('---', 'title: asdf', '---', '', 'Test'), 'test.Rmd')

msg_render = NULL
withCallingHandlers(rmarkdown::render('test.Rmd'), message = function(e) {
  if (grepl('^\nOutput created: ', e$message)) {
    # capture and save the message here, then muffle it
    msg_render <<- e
    invokeRestart("muffleMessage")
  }
})
if (!is.null(msg_render)) message(msg_render)
```

Of course, you can modify `msg_render` before actually emitting the message (e.g., you can tweak the file path contained in the message if necessary).

Does this approach sound okay to you?

cderv:
> I have wished to have a way to suppress only the "Output created" message in the past, too, but I hesitated to add one more argument to rmarkdown::render() just for this purpose. 

By "options", I meant an R option not a new argument to `render()` - something like `options(rmarkdown.rstudio.preview = FALSE)` that we would look for before emitting the message or not. It seems easy to set this option before calling `rmarkdown::render()` from another package. It is one more option though... 

I did not though about handling the message. That is indeed a very good way of doing it! I'll try that. Thank you !

cderv:
@yihui this is now working as expected. 

Thanks for your idea - it made completely sense and reminded me of `suppressMessage`. 

As this is more advanced, I am just letting you check the implementation before merging. Hope that is ok!

yihui (approved):
> By "options", I meant an R option not a new argument to `render()` - something like `options(rmarkdown.rstudio.preview = FALSE)` that we would look for before emitting the message or not. It seems easy to set this option before calling `rmarkdown::render()` from another package. It is one more option though...

I see. I'm okay with adding a global option. That seems to be cleaner than the `withCallingHandlers()` method, so I'd probably do that instead. As you mentioned, it will also allow other packages to suppress the message. We've run into this problem not only in **bookdown**.

cderv:
ok. I suggest we merge this as-is because it is working now and it won't require a new rmarkdown version yet for bookdown.
 
I'll do the change in rmarkdown and we'll change it here when rmarkdown will be on CRAN. 
Does that sound good ?

yihui:
Yes.

PR #1127, opened 2021-04-05
Change devtools for remotes

prdm0:
Encouraging the use of the `remotes` package to install packages kept on github.


yihui (approved):
Thanks!

PR #1131, opened 2021-04-08
Sharing on whatsapp

prdm0:
Informing that it is now possible to share via WhatsApp.

### DIFF
--- a/inst/examples/03-formats.Rmd
+++ b/inst/examples/03-formats.Rmd
@@ -102,6 +102,7 @@ bookdown::gitbook:
       family: sans
       size: 2
     sharing:
+      whatsapp: yes
       facebook: yes
       github: no
       twitter: yes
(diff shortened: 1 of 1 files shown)


cderv:
Hi. 

What is the aim of the PR ? 

I set to false so that isn't activated by default in project. 
Only if users wants to. 

The documentation reflects the default. 

Is it not ok ?

inst/examples/03-formats.Rmd
```diff
@@ -110,7 +111,7 @@ bookdown::gitbook:
       instapaper: no
       vk: no
       whatsapp: no
```
cderv, line 112: See here. I added already.

prdm0:
Hello @cderv, I just want to make it clear to people that sharing on Whatsapp with one click is possible. Several students asked how they would share on Whatsapp. Perhaps new readers, seeing this, will quickly understand that this is possible. Whatsapp is used a lot here and several other places.

Feel free to reject my pull.

Thanks.

prdm0:
Perhaps in your R Markdown Cookbook, you could quote.

cderv:
I already added whatapps to the documentation. So now it is there  twiced 😅

So I have nothing against your PR. It is just already there on the doc.

inst/examples/03-formats.Rmd
prdm0: Oww, sorry, True. I need to stop programming for today. =)

cderv: No worries! 
Thanks for your ideas and suggestions. They are all really welcome !

Issue #585, opened 2018-05-30
`render_book(output_format = "pdf_book")` fails for bookdown-demo when `new_session: yes`


martinjhnhadley:
The discussion of M-K vs K-M rendering options (https://bookdown.org/yihui/bookdown/new-session.html) does not appear to argue against using K-M for generating PDF.

However, the workflow described below yields this error

```r
> render_book(input = c("01-intro.Rmd", "02-literature.Rmd"),
+             output_format = "pdf_book")
Error in eval(parse(text = name)) : object 'pdf_book' not found
Calls: local ... <Anonymous> -> create_output_format -> eval -> eval
Execution halted
Error in Rscript_render(f, render_args, render_meta) : 
  Failed to compile index.Rmd
```

If this workflow is not recommended then I'll happily make a PR to suggest explanatory text to this extent.

# Workflow

1. Download https://github.com/rstudio/bookdown-demo
1. Open bookdown-demo.Rproj in RStudio
1. Add the following to `_bookdown.yml`
`new_session: yes`
1. Run:
```r
bookdown::render_book(input = c("01-intro.Rmd", "02-literature.Rmd"),
            output_format = "pdf_book")
# Error in eval(parse(text = name)) : object 'pdf_book' not found
# Calls: local ... <Anonymous> -> create_output_format -> eval -> eval
# Execution halted
# Error in Rscript_render(f, render_args, render_meta) : 
#  Failed to compile index.Rmd
```

yihui:
It is `bookdown::pdf_book`.

Reference: https://bookdown.org/yihui/rmarkdown/output-formats.html

martinjhnhadley:
Ahh, thanks! I'm not sure how exactly I missed this information.

I added a PR to the demo book here https://github.com/rstudio/bookdown-demo/pull/31

PR #1154, opened 2021-05-14
Fix error in render_book() example

DavidLukeThiessen:
`render_book("index.Rmd",  "pdf_book")` gives an error when ran. According to https://bookdown.org/yihui/rmarkdown/output-formats.html and rstudio/bookdown#585 the output format must specify the bookdown package, which does run correctly.


yihui (approved):
Good catch. Thank you!

PR #1156, opened 2021-05-19
Add display block CSS to figure anchor span

### LINKED ISSUE #1155: bookdown cross-reference links pointing below figures
In reports generated using bookdown, the cross-reference hyperlinks pointing toward Figures seem to point below the Figures (roughly at the caption level), meaning that when clicking on the hyperlink one ends up having only the caption visible on the top of the screen and the reste of the Figure is out of the screen.

This issue only occurs with some browsers such as Chrome or Edge, but not with Firefox, Safari, Internet Explorer or the RStudio viewer. It occurs with both Windows 10 and Android 11 OS.

This message was initially posted on both stackoverflow and RStudio Community :
https://stackoverflow.com/questions/67298237/bookdown-cross-reference-links-pointing-below-figures
https://community.rstudio.com/t/bookdown-cross-reference-links-pointing-below-figures/104394

cderv:
This fixes #1155 

It seems this is required so that the anchor span stays correctly right after the div of class figure. 

@yihui do you have previous experience with this ? 

## tested with 

The link correctly send to the top of the header

````markdown
---
title: "book"
output: 
  bookdown::html_document2: default
---

```{r, echo = FALSE, results='asis'}
cat(stringi::stri_rand_lipsum("10"), sep = "\n\n")
```

```{r fig1, fig.cap = "hello"}
plot(mtcars)
```

See \@ref(fig:fig1)

```{r, echo = FALSE, results='asis'}
cat(stringi::stri_rand_lipsum("10"), sep = "\n\n")
```
````

### DIFF
--- a/R/html.R
+++ b/R/html.R
@@ -665,7 +665,7 @@ parse_fig_labels = function(content, global = FALSE) {
       }
       labs[[i]] = label_prefix(type, sep = ': ')(num)
       k = max(figs[figs <= i])
-      content[k] = paste(c(content[k], sprintf('<span id="%s"></span>', lab)), collapse = '')
+      content[k] = paste(c(content[k], sprintf('<span style="display:block;" id="%s"></span>', lab)), collapse = '')
     }, tab = {
       if (length(grep('^\\s*<caption', content[i - 0:1])) == 0) next
       labs[[i]] = sprintf(

### CODE REVIEW

File: R/html.R [hunk 1]
```diff
@@ -665,7 +665,7 @@ parse_fig_labels = function(content, global = FALSE) {
       }
       labs[[i]] = label_prefix(type, sep = ': ')(num)
       k = max(figs[figs <= i])
-      content[k] = paste(c(content[k], sprintf('<span id="%s"></span>', lab)), collapse = '')
+      content[k] = paste(c(content[k], sprintf('<span style="display:block;" id="%s"></span>', lab)), collapse = '')
```
yihui, line 668: I'm wondering if it's okay to use `div` instead:

```suggestion
      content[k] = paste(c(content[k], sprintf('<div id="%s"></div>', lab)), collapse = '')
```

yihui (approved):
Thanks!

R/html.R
cderv: I have seen some example using a span for other issues(e.g https://stackoverflow.com/a/29123823/3436535)
I don't know if using a div can break something as it is stronger than span for HTML structure, isn't it ? 

Also, changing it would break some custom CSS if any user has used a rule based on the span, but it seems unlikely (I don't see why it would be useful)

Question: idealy, adding the `id` on the `img` tag would be the best, wouldn't it ?  or on the `div` of class figure ? 

Anyway, I think a `div` could work yes. I did not thought of that. 

maybe with `visibility = "hidden"` ? or anything that could help for Accessibility - I wonder how this empty tag renders.

yihui: > So it is safer for now to add a special div or span for our id to avoid conflict right ?

That's what I think.

> We may think more about this if we reimplement the referencing feature based on Lua filter or else.

My impression is that they already add the id to the figure environment instead of using an invisible anchor span, but I could be wrong (you can double check).

> So let's go with a div ? I am fine with it.

Yes, we can revert back to a block span later if we discover any problems.

cderv: > My impression is that they already add the id to the figure environment instead of using an invisible anchor span, but I could be wrong (you can double check).

They put the id to the img tag if you add an id using link attributes
````powershell
❯ pandoc -t html4
![cap](https://rmarkdown.rstudio.com/docs/reference/figures/logo.png){#id}
^Z
<div class="figure">
<img src="https://rmarkdown.rstudio.com/docs/reference/figures/logo.png" id="id" alt="" />
<p class="caption">cap</p>
</div>
````

yihui: Okay I see. Thanks!

cderv:
@yihui I have checked what Quarto does. 

If fact, it adds the id on a wrapping div. 
````html
<div id="fig:script" class="quarto-figure quarto-figure-center anchored">
	  <figure class="figure">
		    <p><img src="images/hopr_0107.png" class="img-fluid figure-img"></p>
		    <p></p>
		    <figcaption aria-hidden="true" class="figure-caption">Figure 1.7: When you open an R Script (File &gt; New File &gt; R Script in the menu bar), RStudio creates a fourth pane above the console where you can write and edit your code.</figcaption>
		    <p></p>
	  </figure>
	  <a class="anchorjs-link " aria-label="Anchor" data-anchorjs-icon="" href="#fig:script" style="font: 1em / 1 anchorjs-icons; padding-left: 0.375em;"></a>
</div>
````

It surely does this for other feature related to classes but this would be another solution for this.

PR #1173, opened 2021-06-10
Document new custom environment syntax

### LINKED ISSUE #939: Document custom block in bookdown book using pandoc fenced Div block
The current edition of the bookdown book currently document using the `block` and `block2` engine. In rmarkdown cookbook we document using pandoc fenced Div block. https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html

We could document the same in bookdown book as this is the new advice way

for a second edition maybe, and for now adding a reference link to Rmarkdown cookbook ? 

Just opening this so we don't forget.

### LINKED ISSUE #1169: No paragraph environments for new paragraphs in math components
This is a screenshot from the _Authoring books with R markdown_: 

image

There is an extra line break in front of each `<span>`, it seems a missing `<p>` for the paragraph causes this glitch.

cderv:
closes #1169 (and other related issues)

This modifies the documentation about Theorem and Proof env using the new fenced Div syntax. 

I did an update of the existing part, and added a sub section about how to convert from old to new syntax. This allow to keep a document of the old syntax but some things where removed (like the need of `echo = TRUE` on these chunk). Should we kept a subsection on this engine syntax ? 

I also added the option to deactivate insertion of default environment definition in preamble. 

At last, currently, we don't have support yet for EPUB or other formats so I tried to emphasized that. The

### DIFF
--- a/inst/examples/02-components.Rmd
+++ b/inst/examples/02-components.Rmd
@@ -230,20 +230,13 @@ Theorems\index{theorem} and proofs are commonly used in articles and books in ma
 In **bookdown**, the types of theorem environments supported are in Table \@ref(tab:theorem-envs). To write a theorem, you can use the syntax below:
 
 ````markdown
-::: theorem
-This is a `theorem` environment which can contain **any** _markdown_ syntax
-:::
-````
-
-or the longer version
-
-````markdown
 ::: {.theorem}
-This is a second theorem
+This is a `theorem` environment that can contain **any**
+_Markdown_ syntax.
 :::
 ````
 
-This syntax is based on Pandoc's [fenced `Div` blocks](https://pandoc.org/MANUAL.html#divs-and-spans) and can already be used in any R Markdown document to write [Custom Block](https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html). **bookdown** is only offering special handling for Theorems and Proofs environments. One important advantage of this syntax is that it is Pandoc's Markdown and you can write any valid markdown inside the block.
+This syntax is based on Pandoc's [fenced `Div` blocks](https://pandoc.org/MANUAL.html#divs-and-spans) and can already be used in any R Markdown document to write [custom blocks.](https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html) **Bookdown** only offers special handling for theorem and proof environments. Since this uses the syntax of Pandoc's Markdown, you can write any valid Markdown text inside the block.
 
 (ref:theorem-envs) Theorem environments in **bookdown**.
 
@@ -256,9 +249,9 @@ knitr::kable(data.frame(
 ), caption = '(ref:theorem-envs)', booktabs = TRUE)
 ```
 
-To write other theorem environments, replace `::: theorem` with other environment names in Table \@ref(tab:theorem-envs), e.g., `::: lemma` .
+To write other theorem environments, replace `::: {.theorem}` with other environment names in Table \@ref(tab:theorem-envs), e.g., `::: {.lemma}`.
 
-A theorem can have a `name` attribute so its name will be printed. In this case, the long version syntax must be used, e.g.,
+A theorem can have a `name` attribute so its name will be printed. For example,
 
 ````markdown
 ::: {.theorem name="Pythagorean theorem"}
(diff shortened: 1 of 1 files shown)


inst/examples/02-components.Rmd
```diff
@@ -230,11 +230,21 @@ Theorems\index{theorem} and proofs are commonly used in articles and books in ma
 In **bookdown**, the types of theorem environments supported are in Table \@ref(tab:theorem-envs). To write a theorem, you can use the syntax below:
 
 ````markdown
-`r ''````{theorem}
-Here is my theorem.
-```
+::: theorem
+This is a `theorem` environment which can contain **any** _markdown_ syntax
```
yihui: For code blocks, make sure the line width is less or equal to 69, otherwise it will exceed the page margin.

inst/examples/02-components.Rmd
```diff
@@ -230,11 +230,21 @@ Theorems\index{theorem} and proofs are commonly used in articles and books in ma
...
+:::
 ````
 
+or the longer version
```
yihui: I think it is very likely that users need other attributes in `{}`, so I tend not to introduce the shorter version `::: theorem` specifically. The longer version is not significantly longer (three more characters), and users will probably always use this syntax. I hope it makes sense to introduce only one syntax here. Otherwise we'll have to repeatedly remind readers that we are using the short or long syntax later.

inst/examples/02-components.Rmd
```diff
@@ -230,11 +230,21 @@ Theorems\index{theorem} and proofs are commonly used in articles and books in ma
...
+
+````markdown
+::: {.theorem}
+This is a second theorem
+:::
+````
+
+This syntax is based on Pandoc's [fenced `Div` blocks](https://pandoc.org/MANUAL.html#divs-and-spans) and can already be used in any R Markdown document to write [Custom Block](https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html). **bookdown** is only offering special handling for Theorems and Proofs environments. One important advantage of this syntax is that it is Pandoc's Markdown and you can write any valid markdown inside the block.
```
yihui: Punctuations need to be inside `[]` because links to be converted to footnotes in PDF, and we want footnote numbers to appear after (instead of before) punctuation marks.

yihui (approved):
Just finished tweaking the text, with a few minor comments (which have been taken care of). Thanks!

inst/examples/02-components.Rmd
cderv: Oh ok. That is specific. 69. I'll try to remember that. Thanks !

inst/examples/02-components.Rmd
cderv: I hesitated about that so I understand the comment. The short version is nice when you do not have any attributes or other class. But yes it may not be useful to introduce two versions, you're right.

inst/examples/02-components.Rmd
cderv: Oh sure. Thanks for the feedback!

inst/examples/02-components.Rmd
yihui: It depends on the font size and typeface. 69 is for this book. In general, I'd try to stay within 60 characters if possible.

inst/examples/02-components.Rmd
yihui: My experience is that when people write a theorem, they are most likely to reference it somewhere (this is too common in math/stats/theory papers), which means they will need to assign an ID to it, hence they will need at least one attribute.

inst/examples/02-components.Rmd
yihui: I remember Pandoc has an option to do that automatically, but I've never used it (perhaps we should).

cderv: I don't think I know this option that would handle the `.`. Or are you meaning that we don't use Pandoc itself to transform links to footnotes ? I see the `links-as-notes` variables is not used in `index.Rmd`. 
How does the transformation works currently ?

yihui: I just looked it up in the manual. I think I was referring to the `notes-after-punctuation` option: https://pandoc.org/MANUAL.html#other-relevant-metadata-fields but you are right that since we didn't use `links-as-notes`, this option probably won't work. Even if we use the `link-as-notes` option, I don't know if `notes-after-punctuation` will work, since it seems to be designed for citations.

cderv: I did not knew that ones - so many extensions! I understand the same : It seems designed for citations anyway.

PR #1174, opened 2021-06-10
[DOCS] Add a box about new custom block syntax in the cookbook

### LINKED ISSUE #939: Document custom block in bookdown book using pandoc fenced Div block

cderv:
This close #939 (I used the incorrect number in other PR #1173)

I thought about this and decided that it was not a good idea to re-document custom block in this book. It would be a duplication in two places. 

So I added a box to mention that the `block` and `bloc2` engine are no more to be used and one should look at the R Markdown Cookbook

Is that enough ? 

We could later just remove this chapter from **bookdown** for a second edition as there would not be any specificity.

### DIFF
--- a/inst/examples/02-components.Rmd
+++ b/inst/examples/02-components.Rmd
@@ -596,6 +596,10 @@ Cross-references still work even when we refer to an item that is not on the cur
 
 ## Custom blocks
 
+:::{.rmdimportant latex=true}
+A new syntax is now adviced to create custom blocks. We invite you to refer to [Chapter 9.6 in the R Markdown Cookbook](https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html) [@rmarkdown2020] and no more use the `block` and `block2` engine described below. Those will be deprecated in future version of **bookdown**. Please make the change in your content if you are still using one of this engine directly.
+:::
+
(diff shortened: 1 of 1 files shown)


yihui (approved):
Yes, a note is enough. We could delete this section later. Thanks!

inst/examples/02-components.Rmd
```diff
@@ -596,6 +596,10 @@ Cross-references still work even when we refer to an item that is not on the cur
 
 ## Custom blocks
 
+:::{.rmdimportant latex=true}
+A new syntax is now adviced to create custom blocks. We invite you to refer to [Chapter 9.6 in the R Markdown Cookbook](https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html) [@rmarkdown2020] and no more use the `block` and `block2` engine described below. Those will be deprecated in future version of **bookdown**. Please make the change in your content if you are still using one of this engine directly.
```
yihui, line 600: > A new syntax is now adviced to create custom blocks.

If an active voice can be used, the passive voice would not be used by me :)

If I can use an active voice, I would not use the passive voice.

Just kidding. This is no big deal, but I guess our editor may point it out before we print the 2nd edition of this book.

PR #1175, opened 2021-06-14
`bs4_book()`: Fix alignment of text blocks within rmdnote divs

### LINKED ISSUE #1084: Code block in rmdnote needs margins adjusted
e.g.

Screenshot 2021-02-22 at 7 32 37 AM

Code should occupy full width of block

maelle:
Fix #1084 (if I understood the problem correctly)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.22.6
+Version: 0.22.7
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # CHANGES IN bookdown VERSION 0.23
 
-- In `bs4_book`, math in footnotes is now rendered (@mine-cetinkaya-rundel, #1026)
+- Style change in `bs4_book()` where code block inside callout blocks will have their background fill the whole width of the bordered block (#1175).
+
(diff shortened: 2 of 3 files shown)


cderv (commented):
This will reduce the pre background and code won't be aligned anymore with the outside text

From the issue, 
> Code should occupy full width of block

Do we want either the opposite and get the pre background to be aligned with the other pre background outside of `rmdnote` box ? 

To illustrate the above, with this PR here is what it looks like:
image

Code and text in the box are no more aligned which does not feel right to me.

Is the screenshot posted in #1084 what it should look like ? Because it seems that the link in mastering-shiny already looks like that.

maelle:
> Because it seems that the link in mastering-shiny already looks like that.

Yes I guess I do not understand what the issue is.

cderv:
@maelle see answer in https://github.com/rstudio/bookdown/issues/1084#issuecomment-860937264

> Like how code blocks on a narrow page occupy the full width of the column, **the code block within a note should also occupy the full width (i.e. no gap between left and right side of code block and containing border)**. Does that make sense? It's not easy to describe in words.

So this PR is not yet quite right. Do you have time to change or should I do this ?

cderv:
cc @apreshill as we discussed it yesterday.

maelle:
image

Btw whilst looking at this on different screen sizes it made the need for a copy-button fix (to not make it hide code) look even more important. :grimacing:

cderv:
I think I would keep the code content align with the text content even if we make the background full width 

It would look like this 
image

cderv:
what is your feeling on this one @apreshill ? I'll tweak the implementation following what we decide. 
Thanks @maelle !

apreshill:
I agree @cderv - I tend to prefer keeping the text aligned inside the note box.

cderv:
So background full width and text align it will be! Thanks.

cderv:
This will now look like this
image

apreshill (approved):
LGTM- thanks!

Issue #1146, opened 2021-04-27
[FR] Add `split_bib` to `bs4_book`


shirdekel:
At the moment the `bookdown::bs4_book` output does not use a `split_bib` argument, like `bookdown::gitbook`. However, I noticed that it seems to be [`FALSE` by default](https://github.com/rstudio/bookdown/blob/2d692ba4b61f1e466c92e78fd712b0ab08c11d31/R/bs4_book.R#L115). Is there a reason why the user can't be given control of this through an argument in `bookdown::bs4_book`? I tried this out in my fork and it seems to work fine, so happy to submit a PR.

sstoeckl:
I would very much like that feature! Could you submit the PR?

shirdekel:
Here it is: https://github.com/rstudio/bookdown/pull/1185

PR #1185, opened 2021-06-24
Add split_bib to bs4_book

shirdekel:
- From https://github.com/rstudio/bookdown/issues/1146

### DIFF
--- a/R/bs4_book.R
+++ b/R/bs4_book.R
@@ -38,7 +38,7 @@
 #' @param repo Either link to repository where book is hosted, used to generate
 #'   view source and edit buttons or a list with repository `base` link, default
 #'   `branch`, `subdir` and `icon` (see "Specifying the repository").
-#' @param lib_dir,pandoc_args,extra_dependencies,... Passed on to
+#' @param lib_dir,pandoc_args,extra_dependencies,split_bib... Passed on to
 #'   [rmarkdown::html_document()].
 #'
 #' @section Specifying the repository:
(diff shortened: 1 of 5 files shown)


R/bs4_book.R
```diff
@@ -48,7 +48,8 @@ bs4_book <- function(
                      ...,
                      lib_dir = "libs",
                      pandoc_args = NULL,
-                     extra_dependencies = NULL
+                     extra_dependencies = NULL,
+                     split_bib = split_bib
```
cderv, line 52: This should be FALSE by default right ? 
```suggestion
                     split_bib = FALSE
```

Also when adding an argument, it should be added to the documentation here: 
https://github.com/rstudio/bookdown/pull/1185/files#diff-ff358cd3f877c575281978ccf49a2c8869cdcef59cdba125698dc3e8966a570aL41

cderv (changes requested):
This seems like a simple change. 

Did you test it against your book ? Did you find any conflict with other features ? 

Can you add a NEWS bullet too ? 

As you added a new argument, you need to document so the CI tests do not error on doc mismatch. thanks !

shirdekel:
Looks like there's still an issue with the documentation. I tried looking at the `gitbook.R` docs but it's not documented there, only in `html.R`.  Should I give it its own `@param` description as in `html.R`?

cderv:
Ok I'll have a look at the issue and do it. Thanks !

cderv (approved):
I have made the change. `split_bib = TRUE` will now be possible to specify in `bs4_book()`. 

This is useful when no CSL with footnote style for citations is used. 
https://pkgs.rstudio.com/bookdown/reference/bs4_book.html#references-bibliography
In that case, `split_bib = TRUE` will do nothing.

When a CSL with footnote style is used, citations are set inline in a linked popup showing on click. This has the same intent a putting the references at the end of the chapter but using a different way to do it.

PR #1186, opened 2021-06-24
Update gha - new testthat output support & remove travis

cderv:
@yihui this PR has two aim: 

* remove travis which is still used for deploying to github pages. I don't think we still need to do that. I removed the yaml and also the tweaks in `_render.R`

* update the R CMD check file to be the same as rmarkdown including the testthat.Rout so that it is read and shown in the report after test. That is helpful to see this.

Hope this is fine with you. I let you check and merge.


yihui (approved):
Sounds good to me. Thanks!

inst/examples/_render.R
```diff
@@ -12,40 +9,36 @@ owd = setwd(dirname(src))
...
-  if (travis) x[i[1]] = r
   i = grep('<i class="fa fa-circle-o-notch fa-spin"></i><a href="./">.+</a>', x)[1]
   # shorter title on the toolbar
   if (!is.na(i)) x[i] = gsub('bookdown: ', '', x[i], fixed = TRUE)
-  i = c(
-    grep('&lt;bytecode: 0x[0-9a-f]+&gt;$', x),
-    grep('^\\s*<meta name="generator" content="bookdown [.0-9]+ and GitBook [.0-9]+" />$', x),
-    grep('^<meta name="date" content="[-0-9]+" />$', x)
-  )
-  if (travis && length(i)) x = x[-i]
```
yihui, line 42: Oh I suddenly recall the reason for publishing to `gh-pages`: I wanted to check the differences in HTML when I switch to a newer version of Pandoc, e.g., https://groups.google.com/g/pandoc-discuss/c/1bwAre8fG6E/m/k4vM455kCQAJ If you feel that's useful, perhaps we need to keep these tweaks and keep pushing to `gh-pages`.

cderv: Oh I see. So the that is why when looking a the webpage `github.io` url, it would redirect to bookdown, but you still could look at the gh-pages branch to see the change. 

That sounds useful. But we currently are build the gh-page branch on every commit to master and I did not see why. It seemed too much to run a CI for nothing.

So now it is not nothing as I understand. But I know I have never looked at the branch `gh-pages` 😅 
Would it be useful to have something we run manually when we want to test a Pandoc version ? 
Or do you look at this branch sometime ? 

Currently, we build the gitbook book on PR, but we can't see the change. Only the book folder can be downloaded on error in the build. 

I am ok to keep it as I see how it can be useful, but maybe I should do it in GHA too to move from travis completly. And maybe we don't need to run it on each commit ? Manual trigger maybe ?

yihui: > Would it be useful to have something we run manually when we want to test a Pandoc version ?

Yes, if we have that, the `gh-pages` branch will be useless and we can remove it.

yihui:
Sounds good! Thanks!

I also started playing with `gh` recently, and it was very convenient indeed!

Issue #1040, opened 2020-12-06
Less obtrusive copy to clipboards on bs4_book


keithmcnulty:
Might be a nicer UX for the copy to clipboard buttons to only appear on `sourceCode` blocks and also for them to only appear on mouseover.  I achieved this with simple custom CSS, but may be worth considering as default?

```css
pre .copy button {
  visibility: hidden;
}

div.sourceCode:hover > pre .copy button {
  visibility: visible;
}
```

Also, would be nicer for the copy button to remain fixed when code blocks are scrolled, similar to gitbook.  Haven't yet worked out the custom css to make that happen.

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.org/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('bookdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/bookdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

keithmcnulty:
Address with #1041

cderv:
Just for later reference as I am looking into this again, `bs4_book()` button is currently placed on any block of code, source and output. 

I believe it was designed with using `collapse = TRUE` knitr option in mind.
However, if FALSE is used, we got this: 
![image](https://user-images.githubusercontent.com/6791940/124279908-51b6e100-db48-11eb-9fd3-d063008a3687.png)

When using an icon, it would maybe be less cluttered #1192

apreshill:
Linking to issue about copy button not showing up on output: https://github.com/rstudio/bookdown/issues/1072 (I agree- do not want)

cderv:
@keithmcnulty we now replaced the copy button by an icon. See https://github.com/rstudio/bookdown/pull/1192#issuecomment-873017992

It is in the dev version. I would be interested in your feedback regarding your obstrusive initial feedback. Thanks!

cderv:
[...]

PR #1041, opened 2020-12-07
Improve copy UX on bs4_book (#1040)

keithmcnulty:
Improves ux of copy to clipboard:

1.  Copy only appears on input code blocks
2.  Only on mouseover
3.  Copy button remains fixed during scroll
4.  Copy button repositioned marginally for better look on single code lines

PR #1187, opened 2021-06-25
bs4_book: no scroll on x overflow and correct copy btn placement

### LINKED ISSUE #1086: Copy copy button fix
https://github.com/r-lib/pkgdown/commit/3574fe0aef12da059875896f5ca21f508aecaefe

maelle:
Fix #1086

Also related #1040 (currently the copy-code button might _hide_ some code)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.22.10
+Version: 0.22.11
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 - In `bs4_book()`, improvement regarding copy button:
   * It has now a light icon instead of a text with white background (#1192). 
   * It will no more show on output block code when knitr's option is `collapse = FALSE` (#1197).
+  * For code block with very long line, it will now wrap in the code block with no more x-scroll bar, and it will wraps around the copy button icon so that text is not hidden (#1187). 
+  If you want to customize part of the UI to change this default behavior, you can do it using a custom css with `bs4_book()`.
 
 - Fix an issue with `bs4_book()` where text written using [Line Block](https://bookdown.org/yihui/rmarkdown-cookbook/indent-text.html) was not found in search (thanks, @dmklotz, #1141).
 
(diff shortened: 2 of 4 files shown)


cderv (approved):
Great. Can you add NEWS bullet so that we track the different changes ? 

I will triage and merge the different bs4_book() items at the end before release to make a single big part on all the changes. Thanks!

maelle:
I'm not sure this should be merged as this can hide text, though?

maelle:
image

cderv:
Thanks. I see. 

This is an issue because we show the copy button always, with no opacity. It would probably no be an issue if the button was shown on hover only (like in `gitbook()`)

This is a topic that we are still thinking about (#1040 and associated PR #1041). 

Best is to deal with this PR when we will deal with the other.

maelle:
Solution I put together with DevTools:

* Put the copy div right at the beginning of the pre node. (prepending instead of appending)
* CSS

```css
pre .copy {
float: right;
margin-left: 1rem;
margin-bottom: 1rem;
}

```

maelle:
image

cderv:
FWIW, quarto as a smaller icon and deal with this by having a padding for the pre block, leaving space between the code part and pre background for the icon 

````css
pre.sourceCode.code-with-copy {
    padding-right: 25px;
}
````
image

This leave the code aligned the same way on the right side.

cderv:
with gitbook() the icon is right above the text in the margin
image

cderv:
@apreshill we should look into that in the same bundle as the copy button improvement, and uniformisation between outputs. 

What are your thoughts on this ?

apreshill:
Hi @cderv looking into this right now. Also another point of comparison is GitHub's new copy icon UI: 

https://github.com/apreshill/bs4booktesting/issues/1

Looks like they actually do the scrolling underneath the button on hover.

I like the padding-right idea, but needed to add 60px to clear the button instead of 25px. Here is how that looks: 
https://apreshill.github.io/bs4booktesting/

Screen Shot 2021-06-30 at 10 30 22 AM


I wish the padding could be a bit more narrow.

---

Using this space to think out loud here:

1. I'd like to keep the existing behavior of permanent copy button (no reveal just on hover). I think that makes sense for a book output, and is more friendly to readers who may not be programmers.
2. @hadley might you be open to switching to an icon here? That would decrease the padding needed. We could follow Quarto's lead and use the Bootstrap clipboard icon: https://icons.getbootstrap.com/icons/clipboard/ (maybe switching to this icon after click: https://icons.getbootstrap.com/icons/clipboard-check/). If you want to keep the word copy, I'll try a few more CSS possibilities.

hadley:
I think it makes sense to match quarto here; I'm fine with the icon.

cderv:
This branch is now tested in https://cderv.github.io/bs4booktesting/wrap-and-scroll/

and it's look like this 
image

Current master looks like this: https://cderv.github.io/bs4booktesting/

image


So this PR: 

* Wraps the content to not have scroll
* make the copy button not hide code by having the text be "around". For reference, the other solution would be a margin for the copy button in the code block : https://github.com/rstudio/bookdown/pull/1187#issuecomment-871607714

@apreshill to discuss.

PR #1192, opened 2021-07-02
bs4_book: Copy button uses an icon instead of text

cderv:
@apreshill as discussed. 

I have used the same logic as Quarto, with icon built in (using inline svg). I have looked at the code there, and reuse part of it

I have kept the style of `bs4_book()` though, with the tooltip and background, and just change the text by an icon, which changes in click for a few seconds.

Example: https://cderv.github.io/bs4booktesting/icon-copy-btn/

We need to deal with the overlap discussed in #1187 (this PR could replace the other or complement only)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # CHANGES IN bookdown VERSION 0.23
 
+- In `bs4_book()`, copy button has now a light icon instead of a text with white background (#1192). 
+
 - Fix an issue with `bs4_book()` where text written using [Line Block](https://bookdown.org/yihui/rmarkdown-cookbook/indent-text.html) was not found in search (thanks, @dmklotz, #1141).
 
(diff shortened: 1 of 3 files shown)


apreshill:
Here is what I'm thinking, will change the css to:

+ remove the border always
+ remove the background color always

I think this will address issue #1040 pretty well.

cderv:
So something like this ? 
image

apreshill (approved):
I really like how this looks and works- thank you! Are you thinking we would tackle the text/icon overlap in a separate PR?

cderv:
> Are you thinking we would tackle the text/icon overlap in a separate PR?

Yes in #1187 after merging this and updating the PR. Keeping things separate in smaller PR seems better to me in order to change the different things on the copy button.

@hadley does this icon and behavior looks ok to you ? 
I prefer to check before merging as this would change the look of your books (which build on bookdown dev version I recall).

hadley:
LGTM

PR #1197, opened 2021-07-06
only apply button to source code block

cderv:
This offer support for `collapse = FALSE` and deal with one of the feedback from #1040

This is also the behavior from `gitbook()`

Testing site: https://cderv.github.io/bs4booktesting/nocopybuttononoutput/#collapse-true-vs-false

Must follow #1187 that should be merged into master then this branch updated


apreshill (approved):
This looks great, thanks!

Issue #1123, opened 2021-04-02
Create `bs4_book` project template


apreshill:
Hi @yihui and @cderv,

In prep for next package release with `bs4_book` output format, we discussed creating a project template with:

1. A project accessible from RStudio project wizard to create  "A Bootstrap-4 HTML book using bookdown"
2. A checkbox for users using GitHub Pages to automatically add the `.nojekyll` file and change the output directory to `docs/` (same as Distill's website, see: https://github.com/rstudio/bookdown/issues/1092; https://github.com/rstudio/bookdown/issues/1087)
3. I'm happy to write some filler text into the book skeleton for this so that the example book gives some minimal instructions for users.
4. An exported function like `create_bs4_book()` to do the same as above from the command line (see: https://github.com/rstudio/bookdown/issues/225); also like Distill's `create_website()` function: https://pkgs.rstudio.com/distill/reference/create_website.html)

cderv:
Thanks for opening so that we don't forget ! 

`bookdown::gitbook()` or `bookdown::bs4_book()` is only format you can change in `_output.yml` in the current existing template. I had in mind adapting and updating the current template.

Would that be worth maintaining two templates ? 

We need at least to add `bs4_book()` to the current template, and work on the `create_book()` function as we discussed (Which is tracked in #225)

What would be the main difference in content than the current gitbook format ? 
What specific content do you see for `bs4_book()` that would not be possible to add in the current template ?

apreshill:
Ah ok- I was thinking of adding a new project template. One of the biggest differences is the one .Rmd per chapter in `bs4_book`, so I was thinking the template would be a great way to make sure that is clear right off the bat. The current skeleton includes a `preamble.tex` file that won't apply, and a CSS file that probably will not work with the bootstrap-based format.

cderv:
> One of the biggest differences is the one .Rmd per chapter in bs4_book, so I was thinking the template would be a great way to make sure that is clear right off the bat. 

`split_by='chapter'` is the default for `gitbook()` too. You just can change it for `gitbook()` and not in `bs4_book()`. The template skeleton currently respect the default, so it would work for both. ([template is here](https://github.com/rstudio/bookdown/tree/master/inst/rstudio/templates/project/resources))

> The current skeleton includes a preamble.tex file that won't apply, and a CSS file that probably will not work with the bootstrap-based format.

[...]

Issue #1092, opened 2021-03-04
Add a checkbox for deploying to ghpages in New Bookdown Project window (inspired by distill)


cderv:
One nice feature of the distill project templates is there is a checkbox for deploying to ghpages, which adds that (`.nojekyll`) for you. It might be nice to add to the bookdown template to help with discoverability:

https://github.com/rstudio/distill/blob/5c62b3d13b4137343ada3c0dc3d2457a796a937c/inst/rstudio/templates/project/website.dcf#L11-L14

_Originally posted by @apreshill in https://github.com/rstudio/bookdown/issues/1087#issuecomment-790879936_

cderv:
We could also add helper functions / checklist to help know what to do depending on the deployment option chosen.

Issue #1087, opened 2021-02-25
Images not rendering on gh-pages; might be missing some config step


lcolladotor:
Hi,

I'm using `bookdown` for https://github.com/lcolladotor/rnaseq_LCG-UNAM_2021 which I'm deplying through `gh-pages`. The images don't seem to be served by GitHub though, so this issue seems related to https://github.com/rstudio/bookdown/issues/133#issuecomment-252422207 and https://community.rstudio.com/t/troubleshooting-broken-images-in-bookdown-hosted-on-gh-pages/38441/6

At is one of the example `<img>`  https://github.com/lcolladotor/rnaseq_LCG-UNAM_2021/blob/ada473a744ebebe1cdcf0fd39b5a42d6c072b631/modelos-estad%C3%ADsticos.html#L470

```
<p><img src="_main_files/figure-html/filter_rse-1.png" width="672" /></p>
```

which does exist at https://github.com/lcolladotor/rnaseq_LCG-UNAM_2021/blob/ada473a744ebebe1cdcf0fd39b5a42d6c072b631/_main_files/figure-html/filter_rse-1.png yet it's not displayed in the final html. It might something to do with GitHub since https://lcolladotor.github.io/rnaseq_LCG-UNAM_2021/_main_files/figure-html/filter_rse-1.png returns a 404 error.

The original R code for generating the figure is https://github.com/lcolladotor/rnaseq_LCG-UNAM_2021/blob/94d0d7e9cc3c03ff0b0f7c38adffda359212ae20/05_modelos.Rmd#L27-L43 with the following YAML https://github.com/lcolladotor/rnaseq_LCG-UNAM_2021/blob/94d0d7e9cc3c03ff0b0f7c38adffda359212ae20/index.Rmd#L2-L14 for configuring the book.

Am I missing something here? Thanks!

Best,
Leo

R Session info: https://lcolladotor.github.io/rnaseq_LCG-UNAM_2021/index.html#r-session-information

cderv:
The issue you encountered is because you did no configure gh-pages with a `.nojekyll` file as documented: https://bookdown.org/yihui/bookdown/github.html

This means you website on gh-pages will be processed by jekyll and so will ignore folder starting with `_`. 
This can be deactivated since 2009 : https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/

As you are using the default book name `_main`, this will create a folder starting with underscore for assets. 
[...]

Issue #225, opened 2016-10-12
Feature request: bookdown.skeleton function


mingsnu:
I think we need a `bookdown.skeleton` function to create a skeleton for a new book. I would be pleased to contribute to this if necessary. Thank you!

yihui:
Similar issue as #116. We will add this feature in the RStudio IDE, and you will be able to create a book project from the menu. For now, you can get started with the bookdown-demo project: https://github.com/rstudio/bookdown-demo

mingsnu:
It would be great to add this feature to RStudio IDE, then I will wait. Cloning the `bookdown-demo` or copying from my own template is what am I doing right now. Thank you:)

yihui:
Actually I think such a function makes sense, and the RStudio IDE can just call such a function to create a book skeleton.

yihui:
You can create a skeleton book project using the RStudio IDE and the development version of bookdown now: https://rstudio.github.io/rstudio-extensions/rstudio_project_templates.html

ColinFay:
Hi @yihui 

I know this issue is closed, yet I still think this could be a nice feature to be able to add a bookdown skeleton to your current directory. 

Here are two use case I've been looking for this feature: 
- Creating a package which includes a bookdown 
- Writing an analysis, which was first supposed to be a plain html, but that I wanted after to turn into a bookdown. 

I know this is just copy and pasting in both case, yet in both these cases it would be more secure to add a skeleton in the current dir. 

This seems like a pretty straightforward thing to code : 
-> Adding current bookdown templates in inst/ 
-> code bookdown::use_bookdown() which would fetch and copy these template to the current directory. 

If you think this is a good idea, I'll be glad to send you a PR with these elements.

yihui:
@ColinFay Have you checked the functions `bookdown::book_skeleton` and `bookdown:::bookdown_skeleton` (N.B. `:::`)? The RStudio IDE uses the latter, and you can use it, too, e.g., `bookdown:::bookdown_skeleton(getwd())`.

krlmlr:
Did you mean to export `book_skeleton()`, and perhaps enalbe generation of `.Rd` documentation for this function? I was looking for exported functions too.

krlmlr:
The initial comment for that function reads that it's not meant to be exported, maybe you could consider this? It seems surprisingly difficult to start a bookdown project when offline -- a link to that function could be featured prominently on `?bookdown::bookdown` .

yihui:
[...]

PR #1201, opened 2021-07-08
Project template with conditional content for gitbook vs bs4_book

apreshill:
Hi @cderv -

Some tweaking will be needed, I decided to put the output format specific resources into separate folders. Not wedded to this approach over others, so feel free to change as you see fit.

I added some minimal pointers inside the chapters themselves for how to do very common/basic things, largely expanding upon what was already there inside the existing intro chapter. In doing so, I realized there is some custom content for `bs4_book` in two .Rmd files:

+ `04-citations.Rmd`
+ `05-blocks.Rmd`

For now, in both of these files, I denoted that content with HTML comments:

`<!--bs4_book-->`
specific things go in here
`<!--end-->`

You might see a better way to handle this.

---
Revisiting our TODO list here from #1123 :

- [x] A project accessible from RStudio project wizard with drop-down (some wiring in this PR already provided by @cderv, needs tweaking)

- [ ] A checkbox for users using GitHub Pages to automatically add the `.nojekyll` file and change the output directory to `docs/` (same as Distill's website, see: https://github.com/rstudio/bookdown/issues/1092; https://github.com/rstudio/bookdown/issues/1087)

- [x] I'm happy to write some filler text into the book skeleton for this so that the example book gives some minimal instructions for users. (this PR right here)

- [x]  Exported functions `create_bs4_book()` and `create_gitbook` to do the same as above from the command line (see: https://github.com/rstudio/bookdown/issues/225); also like Distill's `create_website()` function: https://pkgs.rstudio.com/distill/reference/create_website.html) (done by @cderv)


cderv (commented):
I rework the logic so that it works with folders. 

I also added a mechanism to remove block in magick comment `<!--format:start-->` and `<!--format:start-->`. For the desired format, the comment only are removed, for the other format the whold block in between comments are removed. 

I also refactored everything. Use smaller functions and added some tests for this. 

I think it works ok now. 

I did not yet take into accound the gh-page stuff. And I did not look at the content of the Rmd yet.

apreshill:
Summing up discussion with @cderv and polishing up this PR:

- [ ] Doc updates are needed. The [custom blocks chapter](https://bookdown.org/yihui/bookdown/custom-blocks.html) in bookdown book- needs to essentially be archived. Suggestion: pull attention to `bs4_book` as a way to use a custom `.rmdnote` block, and maybe show styling examples of Pandoc div blocks for `gitbook` (and `bs4_book` also). Decision: include single example of a custom block that works for both PDF and HTML output formats, and would work for gitbook.

Best docs on this for now in (but not specific to bookdown users): https://bookdown.org/yihui/rmarkdown-cookbook/custom-blocks.html

- [ ] In bs4_book template, note that `.rmdnote` won't work for PDF output unless you define it...point to new bookdown section on this.

- [ ] In gitbook template, mention custom blocks with link to new bookdown section on this.

apreshill:
@cderv I'm running into errors using the `create_*` functions. Probably just me but:

```
> create_bs4_book("/Users/alison/rscratch")
Error in file(con, "w") : cannot open the connection
In addition: Warning message:
In file(con, "w") :
  cannot open file '/Users/alison/rscratch/help/index.Rmd': Permission denied
Called from: file(con, "w")
Browse[1]> Q
> create_bs4_book("/Users/alison/rscratch/")
Error in file(con, "w") : cannot open the connection
In addition: Warning message:
In file(con, "w") :
  cannot open file '/Users/alison/rscratch/help/index.Rmd': Permission denied
Called from: file(con, "w")
Browse[1]> 
```

cderv:
I still need to fix a weird issue on windows with the test I have made  to check the book build on a background session. I need to look into that before merging this PR.

cderv:
So CI issue is fixed. It was not easy to find. 

I needed to use `new_session = FALSE` specifically for this skeleton test of `bs4_book()`
There is something very odd happening during the testing when running `render_book()`. Leaving some notes for future me.

<details>
<summary>About the issue</summary>

Issue was: 
````
> test_check("bookdown")
Error: Error in loadNamespace(x) : there is no package called 'bookdown'
Error: Error: R CMD check found ERRORs
Calls: local ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
Execution halted
== Failed tests ================================================================
-- Error (test-skeleton.R:72:3): Created bs4_book template works ---------------
Error: Error: Failed to compile index.Rmd
Backtrace:
    x
 1. +-base::suppressMessages(render_book(dir, quiet = TRUE)) test-skeleton.R:72:2
 2. | \-base::withCallingHandlers(...)
 3. \-bookdown::render_book(dir, quiet = TRUE)
 4.   \-bookdown:::render_new_session(...)
 5.     \-bookdown:::Rscript_render(...)

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 52 ]
Error: Error: Test failures
Execution halted

1 error x | 0 warnings v | 0 notes v
Error: Process completed with exit code 1.
````

So oddly, when using `bookdown::Rscript_render`, the package is not found. 

I can't reproduce this locally, and can't find anymore the why. I recall it has something to do with `readRDS` and `saveRDS` on an object for namespace bookdown. I recall they needed to check on the pkg namespace at some point. 

But I don't know why it can't be found - the package is installed for sure duing R CMD check 🤔 
</details>

This PR is now ready unless you have other tweak to make @apreshill

**Review by cderv — APPROVED** [review]

File: inst/rstudio/templates/project/resources/bs4_book/_output.yml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+bookdown::bs4_book:
+  css: bs4_style.css
+  theme:
+    primary: "#096B72"
+  repo: https://github.com/johndoe/bookdown-demo
```
yihui, line 5: `johndoe` is an existing username. We should probably use one that doesn't exist, e.g.,

```suggestion
  repo: https://github.com/USERNAME/bookdown-demo
```

inst/rstudio/templates/project/resources/common/04-citations.Rmd
```diff
@@ -0,0 +1,20 @@
...
+## Citations
+
+Reference items in your bibliography file(s) using `@key`.
+
+For example, we are using the **bookdown** package [@R-bookdown] (check out the last code chunk in index.Rmd to see how this citation key was added) in this sample book, which was built on top of R Markdown and **knitr** [@xie2015] (this citation was added manually in an external file book.bib). 
+Note that the `.bib` files need to be listed in the index.Rmd with the YAML `bibliography` key.
+
+<!--bs4_book:start-->
+
+The `bs4_book` theme makes footnotes appear inline when you click on them. In this example book, we added `csl: chicago-fullnote-bibliography.csl` to the `index.Rmd` YAML, and include the `.csl` file.
```
yihui, line 16: Is this `.csl` file mandatory to `bs4_book`? I mean, does this feature work without a `.csl` file, and does it work with other `.csl` files?

inst/rstudio/templates/project/resources/common/06-final-words.Rmd
```diff
@@ -0,0 +1,33 @@
...
+
+Read more about the features of this output format here:
+
+https://pkgs.rstudio.com/bookdown/reference/gitbook.html
+
+Or use:
+
+```{r eval=FALSE}
+library(bookdown)
+?gitbook
```
yihui, line 29: In general, I tend not to load a package unless it's really necessary:

```suggestion
?bookdown::gitbook
```

inst/rstudio/templates/project/resources/common/index.Rmd
```diff
@@ -0,0 +1,55 @@
+--- 
+title: "A Minimal Book Example"
+author: "John Doe"
+date: "`r Sys.Date()`"
+site: bookdown::bookdown_site
+url: https://bookdown.org/johndoe/bookdown-demo
+cover-image: cover.png
```
yihui, line 7: We have a copy of the cover image here: https://github.com/rstudio/bookdown/blob/master/inst/examples/images/cover.jpg so I tend not to duplicate it here.

```suggestion
cover-image: https://bookdown.org/yihui/bookdown/images/cover.jpg
```

inst/rstudio/templates/project/resources/common/index.Rmd
```diff
@@ -0,0 +1,55 @@
...
+documentclass: book
+bibliography: [book.bib, packages.bib]
+yaml: goes here
```
yihui, line 10: What's this?

inst/rstudio/templates/project/skeleton.dcf
```diff
@@ -1,4 +1,11 @@
-Title: Book Project using bookdown
-Binding: bookdown_skeleton
-OpenFiles: README.md, _bookdown.yml, _output.yml, index.Rmd
+Title: Book project using bookdown
+OpenFiles: README.md, index.Rmd, _bookdown.yml, _output.yml
```
yihui, line 2: I think I meant to open `index.Rmd` last so it will be the first file in the editor. The order of files were changed here, and I guess users will first see `_output.yml`?

R/skeleton.R
```diff
@@ -1,25 +1,161 @@
...
+  bookdown_skeleton_build_index(path, output_format)
+  bookdown_skeleton_build_output_yml(path, output_format)
+  bookdown_skeleton_build_bookdown_yml(path, output_format)
+  move_dir(file.path(path, output_format), path) # move left format files
+  bookdown_skeleton_remove_blocks(path, output_format)
+
+  invisible(TRUE)
+}
+
+bookdown_skeleton_remove_blocks = function(path, output_format) {
```
yihui, line 29: Except for `bookdown_skeleton`, I'd get rid of the prefix `bookdown_` for the rest of helper functions below, e.g., use `skeleton_remove_blocks` instead of `bookdown_skeleton_remove_blocks`. That's because `bookdown_skeleton` could potentially be exported in the future, and `skeleton` wouldn't be a proper name. The rest of helper functions are highly likely to stay internal forever, so we don't need to prefix them.

yihui (approved):
Thanks! Just some small comments and questions.

inst/rstudio/templates/project/resources/bs4_book/_output.yml
cderv: Oh that is true ! Good thinking ! 

Should we use our own repo otherwise https://github.com/rstudio/bookdown-demo ? 

@apreshill I let you decide.

inst/rstudio/templates/project/resources/common/04-citations.Rmd
cderv: The CSL is not mandatory. The feature is : 

* If citation are in footnote style, then the footnotes are transformed to be show inline 
  image
* If not, then they appear as usual. This is without CSL:
  image
  but currently, the `split_bib` option is hard coded to FALSE (should be opt-in #1185)

So CSL style is mandatory to make the citation appear inline in clickable footnote upperscript number. But this seems ok to do that only if the citation style allows it.

Also, this feature will work with any citation style making them footnotes, so ones from https://www.zotero.org/styles?format=note I guess, but I don't know other popular unless this one. For example, it works also with `the-journal-of-clinical-ethics.csl` (https://www.zotero.org/styles/the-journal-of-clinical-ethics)

Does it answer your questions ?

inst/rstudio/templates/project/resources/common/index.Rmd
cderv: The idea here is to have a template project to start with. 

I am fine with a link here to show that this is possible too, but in our book we usually use a cover image that is in the book, and that gets preprended with the url. See
 * [R Markown Cookbook](https://github.com/yihui/rmarkdown-cookbook/blob/b7d818ccd5ff569194a67d1462a94f85716f5fa8/index.Rmd#L20)
 * [Bookdown Book](https://github.com/rstudio/bookdown/blob/b3f11ba9daed638f3a71e1fecd6c2e626f80d355/inst/examples/index.Rmd#L18)
 * [Definitive Guide](https://github.com/rstudio/rmarkdown-book/blob/ed81b894d5d2665d6ef4657c5840634512b97ff5/index.Rmd#L21)

That is why I thought showing the common practice was a good idea. 

We could maybe not include it and just add a paragraph on how to add a cover image using this field ?

inst/rstudio/templates/project/resources/common/index.Rmd
cderv: This is a placeholder mechanism so that we can insert the non common part of YAML header for index.Rmd between `gitbook()` and `bs4_book()`. 

My idea was to not duplicate so to have: 

* Common elements and format specific elements
* A mechanism to build the template from those two. 

So this placeholder is replaced in `bookdown_skeleton_insert_yml()` here: https://github.com/rstudio/bookdown/pull/1201/files#diff-f9db8366f31625a76f20758c5adc9d482d7a5243cb2ebe2d5cc2c9cd506fd7ccR61
I have done that to not add a dependency on a templating system like **whisker**

Maybe I should make that more clear by commenting or using a "non supported" YAML syntax ? 

```suggestion
<< placeholder >>
```

Do you have other ideas ? 

Or is aiming at not duplicate the common template project component a wrong idea ? 

The other mechanism is with the comment in Rmd files 
````
<!--bs4_book:start-->

Content specific to bs4_book()

<!--bs4_book:end-->
````

File: inst/rstudio/templates/project/skeleton.dcf (hunk 6 above)
Comment (cderv) [reply inst/rstudio/templates/project/skeleton.dcf:2]: They may have changed something but when I try in the IDE stable release (1.4.1717), the first file in the list is the one that end up activated, so README.md

The order is then the order of the file in the source pane.

File: R/skeleton.R (hunk 7 above)
Comment (cderv) [reply R/skeleton.R:29]: First let me explain this try :) 

I added those prefixes to make clear to me (as a developer and maintainer) that those internal functions where related to `bookdown_skeleton` main function. 
I think I got that from **renv** seeing how many small functions there are, and that (probably) all internal functions were prefixed, following a pattern like `renv_<main-topic>_<function_name>`. For example, `renv::dependencies()` => `renv:::renv_dependencies_*()` e.g `renv:::renv_dependencies_discover_r()`

I think I like this style for clarity regarding internal functions for us and I wanted to tried it here :) 

But I have no hard feeling on it. I can surely remove `bookdown_` prefix. About `bookdown_skeleton()`, we also need to keep this function name unchanged because it has been communicated and taught in the past (see https://github.com/rstudio/bookdown/issues/225#issuecomment-494054273) so it is used already ! But I hope `create_*()` will be enough and this `bookdown_skeleton()` should not be exported.

File: inst/rstudio/templates/project/resources/common/04-citations.Rmd (hunk 2 above)
Comment (yihui) [reply inst/rstudio/templates/project/resources/common/04-citations.Rmd:16]: Yes. I prefer not shipping a csl file with this package unless necessary, since shipping it means we'll need to maintain it (although I guess it requires little maintenance).

I'll let you decide.

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 4 above)
Comment (yihui) [reply inst/rstudio/templates/project/resources/common/index.Rmd:7]: > We could maybe not include it and just add a paragraph on how to add a cover image using this field ?

Yes, that's what I prefer.

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 5 above)
Comment (yihui) [reply inst/rstudio/templates/project/resources/common/index.Rmd:10]: Oh I see. I'm fine with `yaml: goes here`. I thought it was a YAML field that was accidentally left over.

Perhaps a comment like this will be more obvious:

```suggestion
# additional yaml goes here
```

inst/rstudio/templates/project/skeleton.dcf
yihui: Okay. That sounds good to me.

R/skeleton.R
yihui: > I added those prefixes to make clear to me (as a developer and maintainer) that those internal functions where related to `bookdown_skeleton` main function.

They all have already had a common prefix `skeleton_`, which I feel is enough to tell that these functions are related to creating a book skeleton. In general, I prefer shorter names without losing much clarity (otherwise it feels like programming in Java :). In this case, I feel `skeleton_foo()` is as clear as `bookdown_skeleton_foo()`. But I'll let you decide---whoever writes the majority of the code has the decision right :)

> But I hope `create_*()` will be enough and this `bookdown_skeleton()` should not be exported.

I hope so, too.

yihui (approved):

inst/rstudio/templates/project/resources/common/04-citations.Rmd
cderv: I think it is important to show this feature that is not common and difficult to guess. 
I see several soluton: 
* We can use a URL and it would work. `csl` field supports that. 
````yaml
csl: https://www.zotero.org/styles/chicago-fullnote-bibliography
````
* We can have a script to maintain external data up-to-date by downloading the source. Either https://www.zotero.org/styles/chicago-fullnote-bibliography or directly for repo https://github.com/citation-style-language/styles/blob/master/chicago-fullnote-bibliography.csl

Maybe the first solution is best ?

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 5 above)
Comment (cderv) [reply inst/rstudio/templates/project/resources/common/index.Rmd:10]: I'll make the change to a comment. That will be more explicit for future us and others

File: R/skeleton.R (hunk 7 above)
Comment (cderv) [reply R/skeleton.R:29]: > I feel skeleton_foo() is as clear as bookdown_skeleton_foo(). 

I agree the `skeleton_*` prefix is enough. I'll change that.

> But I'll let you decide---whoever writes the majority of the code has the decision right :)

😄 Agreeing on some common rules and consensus feels better on the long run ! 😉

File: inst/rstudio/templates/project/resources/common/04-citations.Rmd (hunk 2 above)
Comment (yihui) [reply inst/rstudio/templates/project/resources/common/04-citations.Rmd:16]: The downside of the first solution is that users won't be able to build the book offline. One intermediate solution is that we download it automatically the first time when the book is built, i.e., we add a code chunk to download the csl if it doesn't exist:

````
```{r, download-csl, include=FALSE}
if (!file.exists("chicago-fullnote-bibliography.csl")) xfun::download_file(
  "https://www.zotero.org/styles/chicago-fullnote-bibliography",
  "chicago-fullnote-bibliography.csl"
)
```
````

Comment (cderv) [reply inst/rstudio/templates/project/resources/common/04-citations.Rmd:16]: Oh this is in fact a nice example to show! I like that.

Comment (apreshill) [reply inst/rstudio/templates/project/resources/common/04-citations.Rmd:16]: Will this be in the template book itself? Or can the template function run this code upon book creation? I think providing that code chunk in the template feels like ["inessential weirdness"](https://www.harihareswara.net/sumana/2016/05/21/0). I worry it looks complicated and will alienate users. On the other hand, the inclusion of this file seems integral to how bs4_book works. 

Would you be able to live with including this file, but then in that section of the template me noting that we suggest downloading an up-to-date CSL file from https://www.zotero.org/styles/?

That feels like a very useful link to share, as does a real working example inside the template itself.

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 4 above)
Comment (apreshill) [reply inst/rstudio/templates/project/resources/common/index.Rmd:7]: Could we re-visit this? Can we make the image super small? This is a tricky one (because you have to use `cover-image` PLUS `url`) for it to work properly, and easier to see with a working example than to read.

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 5 above)
Comment (apreshill) [reply inst/rstudio/templates/project/resources/common/index.Rmd:10]: @cderv can you resolve this when you make the change you have in mind?

File: R/skeleton.R (hunk 7 above)
Comment (apreshill) [reply R/skeleton.R:29]: ok to resolve?

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 4 above)
Comment (yihui) [reply inst/rstudio/templates/project/resources/common/index.Rmd:7]: Yes, we could use the same approach as https://github.com/rstudio/bookdown/pull/1201#discussion_r675653417.

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 5 above)
Comment (cderv) [reply inst/rstudio/templates/project/resources/common/index.Rmd]: Change in 125b862

File: R/skeleton.R (hunk 7 above)
Comment (cderv) [reply R/skeleton.R]: change done in e277143

File: inst/rstudio/templates/project/resources/common/04-citations.Rmd (hunk 2 above)
Comment (cderv) [reply inst/rstudio/templates/project/resources/common/04-citations.Rmd:16]: * Add the download function to `create_bs4_book()` and make it work offline (do we have a way to detect offline usage ?)
* Add a not in the template about CSL
* Add a link to zotero website in `?bs4_book()`

File: inst/rstudio/templates/project/resources/common/index.Rmd (hunk 4 above)
Comment (cderv) [reply inst/rstudio/templates/project/resources/common/index.Rmd:7]: * Retrieve a file from within the package instead of creating a new one.

**Comment by apreshill** [comment]
Hi @yihui and @cderv -

I believe I tackled all the edits we discussed this AM:

- [x] Added the `edit` key for gitbook (note to self: the keys are different per format now- current docs for custom configs are not specific to output formats) https://github.com/rstudio/bookdown/pull/1201/commits/704a43fbfbd8d57ccfa1b9bf775ee878512863e3

- [x] Add 404 page and document together with HTML metadata https://github.com/rstudio/bookdown/pull/1201/commits/5e98955f1ef01926bdbb4b5a469fdd9bbd8100b5

- [x] Clarified custom blocks section to link to R Markdown cookbook: https://github.com/rstudio/bookdown/pull/1201/commits/9ed447570d528dcd66bea5e34569f47f2a8ec49f

- [x] Add Zotero link for downloading csl style files: https://github.com/rstudio/bookdown/pull/1201/commits/1fe0a05dcad7a950cd783d6d7575b74dd9e29761

File: inst/rstudio/templates/project/resources/common/04-citations.Rmd (hunk 2 above)
Comment (cderv) [reply inst/rstudio/templates/project/resources/common/04-citations.Rmd]: I have added a function to try download asset. When offline (or rather download is not working), it will print a message

````r
> path = tempfile()
> create_bs4_book(path)
>>> ‘chicago-fullnote-bibliography.csl’ can't be downloaded. Please retrieve it manually at ‘https://www.zotero.org/styles/chicago-fullnote-bibliography’ or remove its usage from the template.
[1] "C:\\Users\\chris\\AppData\\Local\\Temp\\Rtmp2XNeZd\\file18ca04f252c6d"
```

Best way would be to use `curl::has_internet()` / `curl::nslookup()` to test offline or not, but we don't have curl in dependency and I don't want to add it for this only.

With 3a0441e we now have the csl file downloaded when the template is created the first time.

inst/rstudio/templates/project/resources/common/index.Rmd
cderv: We now retrieve the logo. But this is still not great IMO. I'll explain in main thread

cderv:
I have now taken into account all feedback: 

* CSL file is downloaded during `create_bs4_book()` function. I used a `tryCatch` call so that we don't error is download is not possible and the project is still created. 

* `cover.png` has been removed. We now copy the logo included in `man/` which is in `help/` when install. (devtools' `system.file` shims does not account for that). However, this will not be so useful because
  *  `cover-image` can be a filename concatenated to `url` variable.
  * `cover-image` can be absolute url otherwise

As the template is not hosted, provided `cover.png` is not useful because it will be concatenated to our dummy url and it won't be really useful. The only purpose would be to show that this exist, but the fact that `cover.png` exists does not matter. We could use 
````yaml
cover-image: replace-me.png
````
with the same intent to show that the config exist.

I am just not sure at the end if `cover.png` needs to be in the resulting template folder. 

* Repo url in bs4_book() template needs to be set to non-existing repo or one or our repo. 

@yihui @apreshill that is the last thing to do before merging.

**Comment by cderv** [comment]
@apreshill unless you want to add something more about `cover-image` in sketelon Rmd, I'll merge this PR tomorrow.

File: inst/rstudio/templates/project/resources/common/index.Rmd [hunk 8]
```diff
@@ -0,0 +1,55 @@
+--- 
+title: "A Minimal Book Example"
+author: "John Doe"
+date: "`r Sys.Date()`"
+site: bookdown::bookdown_site
+url: https://bookdown.org/username/bookdown-demo
+cover-image: your-social-sharing-image.png
```
yihui, line 7: I feel it would be better to have a working example here like:

```suggestion
# Please update the info below according to the actual url
# and cover-image of your book. If you do not have a cover
# image, you may delete url and cover-image below.
url: https://bookdown.org/yihui/bookdown
cover-image: images/cover.jpg
```

otherwise I prefer commenting out these fields:

```yaml
# url: your book url like https://bookdown.org/yihui/bookdown
# cover-image: path to the cover image like images/cover.jpg
```

I'm okay with either way.

yihui (approved):
Looks good! Thanks!

inst/rstudio/templates/project/resources/common/index.Rmd
apreshill: I preferred the second option here, so I made this change.

apreshill:
@cderv I believe I am ready to merge 👍

PR #1203, opened 2021-07-10
[Book] Add link that explains 80/20 rule

salim-b:
(not all people/cultures might be familiar with this term)


cderv:
Thank you !

PR #1204, opened 2021-07-12
Revert and improve change on code block regarding scrolling

cderv:
This reverts commit dc872256e297ee1ce8e451b42d65b53fc8e3720d (#1187) following discussion in #1040

Plan is to keep the scrolling but have the icon stays on the top right side even when scrolling. 

The button icon will always show so there could be overlap with long line as we no more wrapp. However, the icon is a lighter option than a button, and by default there is an opacity filter. It will be darker on button hover only.

Demo here: https://cderv.github.io/bs4booktesting/with-scroll/

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.22.11
+Version: 0.22.12
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.22.10
+Version: 0.22.11
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.22.11
+Version: 0.22.10
 Authors@R: c(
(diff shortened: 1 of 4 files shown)


apreshill (commented):
This looks good. Two minor suggestions:

1. I don't think the margins are needed here anymore now (red background color added here to help visualize):
   Screen Shot 2021-07-12 at 11 43 43 AM
1. I have a minor preference for 50% opacity 😎 

   https://user-images.githubusercontent.com/12160301/125339994-c509e100-e306-11eb-9187-7798e13d7990.mov

cderv:
Ok for the 50%. It looks ok ! 

For the margin around the button I thought I removed them. I'll look into it. Maybe bootstrap add some by default we need to reduce. 
Does it cause any issue in layout having the red margins above ?

apreshill:
I don't mind the padding as is, as long as it is symmetrical. The way the margins are added currently (on top of the padding), there is more space on the right than the top.

Quarto is unbalanced, which bothers me!
Screen Shot 2021-07-12 at 3 24 23 PM

I would do `.5rem` all around.

cderv:
> I don't think the margins are needed here anymore now (red background color added here to help visualize):

I know understand you are talking about the margin of the pre, is that it ? 
I am seeing that in the screenshot looking at what rule you deactivated.

The current margin on the pre create a gap between the button and the right side of the pre block when you add the red background. 
This margin on the pre is there by designed I believe to have the background of code block spread completely from left to right. (Which lead also to change in block for https://github.com/rstudio/bookdown/issues/1084

Currently, `<button>` have no margins and padding is the rule sets by bootstrap. 
image

> I don't mind the padding as is, as long as it is symmetrical.

What we could do is move the copy button to the right to be coherent with the pre background. 

image

cderv:
If we want to reduce the red background padding, it would mean changing bootstrap default. 

Possibly this would need to be done at the bootstrap variable level maybe ? 

* rule for `.btn` using `button_size()` mixin on `$padding-base-vertical` and `$padding-base-horizontal`
https://github.com/rstudio/bslib/blob/0650d9d8dec17cb81fafc351efe2ec11da1ac16e/inst/lib/bs3/assets/stylesheets/bootstrap/_buttons.scss#L20

* `button_size()` sets the padding with the above values
https://github.com/rstudio/bslib/blob/0650d9d8dec17cb81fafc351efe2ec11da1ac16e/inst/lib/bs4/scss/mixins/_buttons.scss#L105

I am tweaking a lot of things at CSS level currently, but we should leverage bslib better maybe. the `bs4_book.css` could have been a `bs4_book.scss` at some point. 😄

cderv:
Updated example: https://cderv.github.io/bs4booktesting/with-scroll/final-words.html

apreshill (approved):
LGTM!

PR #1211, opened 2021-07-28
Pin Boostrap version to 4 in bs4_book theme

cderv:
cc @apreshill 

To do different tweak we think are necessary regarding what is possible for theming

- [x] Pin version 4 of bootstrap: This will prevent any side effect if bslib decide to change its default. 
  Could still be changed with 
  ````yaml
  bookdown::bs4_book:
    theme:
      version: 5
  ````
  if a user really want to

### DIFF
--- a/R/bs4_book.R
+++ b/R/bs4_book.R
@@ -192,9 +192,9 @@ bs4_book <- function(theme = bs4_book_theme(),
 #' @export
 #' @rdname bs4_book
 #' @param primary Primary colour: used for links and background of footer.
-#' @param version Passed to `bslib::bs_theme()`. This should not be changed as
-#'   `bs4_book()` has been designed to work with Bootstrap version 4 for now.
-#' @param
+#' @param version Passed to [bslib::bs_theme()]. This should not be changed as
+#'  [bs4_book()] has been designed to work with Bootstrap version 4 for now.
+#' @md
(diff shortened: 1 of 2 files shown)


R/bs4_book.R
```diff
@@ -192,8 +192,9 @@ bs4_book <- function(theme = bs4_book_theme(),
 #' @export
 #' @rdname bs4_book
 #' @param primary Primary colour: used for links and background of footer.
-bs4_book_theme <- function(primary = "#0068D9", ...) {
+bs4_book_theme <- function(primary = "#0068D9", version = 4, ...) {
   bslib::bs_theme(...,
+    version = 4,
```
cderv, line 197: In fact it could be better to just enforce version 4 as you tried and it won't work with 5. And the book is called `bs4_book()` :) 

Better or not ?

apreshill: So the first solution allows the manual override, but the second does not? I could see going either way, but leaning towards allowing overriding and just calling that clearly off label use (i.e., all warranties/docs voided)? What do you prefer?

cderv: Yes exactly, that is the difference. 

Let's make it clear that `version` should not be changed then.

apreshill: do you mean adding to the docs?

cderv: Yes to the function doc.

PR #1217, opened 2021-07-28
add netlify continuous deployment section

apreshill:
This PR adds a section on Netlify continuous deployment, right after a section on Netlify Drop, to the publishing chapter of the book.


cderv (commented):
A few comments below because I don't understand some stuff. 

Also, I wonder if it would be better to have a Netlify chapter, then two subchapter , one one Drop and one on deployement

````
6. Publishing
    |>> RStudio Connect
    |>> Netlify
        | >> Netlify Drop
        | >> Netlify Deployment 
    |>> Github
        |>> Github Pages
        |>> Using Github Actions
````

I understand you do that to offer option from simpler to more advanced, right ? 

Maybe the publishing section needs further restructuring later on to organize by type of method or other instead of services. 🤔

**Comment by apreshill** [comment]
Re: the organization. What you propose was how I did have it, but then the subsections end up heavily nested. And folks who use Netlify won't find the GitHub Actions section. What I would do:

```
6. Publishing
    |>> RStudio Connect
    |>> Netlify Drop (no CI)
    |>> Netlify Continuous Deployment (no CI)
    |>> Github Pages (no CI, to `docs/`) _[this and the above section could be fused- maybe?]_
    |>> Continuous integration (Github Actions to deploy to branch, maybe leave off Travis)
           |>> show both Netlify + GH Pages 
```

cderv:
Yes I agree I think this makes sense. I would add rstudio connect on the CI part because this is what we use for our books and it is working quite well. (However, it requires maybe some simplification - needs to find time to document and update `publish_book()` function)

yihui (commented):
Like https://github.com/rstudio/bookdown/pull/1218#pullrequestreview-722793171, I feel both the documentation and usage will become easier after we provide an action or a function `publish_netlify()`. Users would only need to provide a site id and personal access token, and we can take care of the rest of technical details.

apreshill:
I would LOVE to document that ;)

Issue #1190, opened 2021-06-30
`bookdown::clean_book()` should ignore `.nojekyll` file used when deployed on github pages.


cderv:
Scenario: 

* Use `docs/` as `output_dir`
* Build the book
* Manually adds `.nojekyll` inside the folder (related to https://github.com/rstudio/bookdown/issues/1092 for better config)
* Push to Github
* Publish `docs/` as github pages

Cleaning the project should not remove the manually created jekyll file. This could be an option in `clean_book(clean = TRUE, keep_nojekyll = TRUE)`

cderv:
Using 

```yaml
resource_files:
  - .nojekyll
```
and `file.create(".nojekyll")` does not help. 

For now, workaround is to set 
````r
file.create(bookdown:::output_path(".nojekyll"))
````

in a setup chunk. 

So we could also offer a setup function to run for this 

````r
bookdown::use_github_pages()
````

that would do this 

Or just export `output_path` and document.

PR #1218, opened 2021-07-29
Update GitHub Pages section

apreshill:
This PR updates the GitHub Pages section (and splits off the existing section on Travis). Cl topics like Travis/GH actions remain to be updated.


inst/examples/06-publishing.Rmd
```diff
@@ -19,42 +19,103 @@ bookdown::publish_book(render = 'local')
...
+You can host your book on GitHub\index{GitHub} for free via GitHub Pages (<https://pages.github.com>).^[See the [GitHub Pages Help documents](http://bit.ly/2cvloKV) for more information.] The simplest approach is to publish your book as a [GitHub Project Page](https://help.github.com/en/articles/user-organization-and-project-pages#project-pages-sites) from a `/docs` folder on your `main` branch, which will then be available at `https://<USER>.github.io/<REPO>` (unless you provide a custom domain name). The main benefit to this approach is its simplicity, and the fact that you can track the source files for your book and the published HTML files in the same branch.
```
cderv, line 24: Is this the preferred way all the time compared to `gh-pages` ? 
Or is this what you think is simpler to use with Bookdown ? 

My view on this: 
* `/docs` folder is simpler when not using a CI: 
   * Build locally to `docs/`
   * Commit and push

* I find `gh-pages` cleaner to have in the source repo and easier to put in place when using a CI. Locally I use a Makefile to help me generally 
  * On main branch, `_book` (or other output dir) is in `.gitignore`
  * I build the book locally 
  * Then I switch branch to gh-pages
  * I commit the `_book` folder
  * I push the gh-pages branch 
  * I switched back to main branch

Makefile do all that for me usually. Using a R file with **gert** should use the same. An R function could help with that.

Anyway, just sharing how I work with this.

inst/examples/06-publishing.Rmd
```diff
@@ -19,42 +19,103 @@ bookdown::publish_book(render = 'local')
...
+
+Create a hidden, empty file named `.nojekyll` from within R in your **bookdown** project root directory:
+
+```
+file.create(".nojekyll")
+```
+
+If you are using RStudio, you may need to refresh the viewer pane to see this new file. You can also use the terminal to create this file with the command `touch .nojekyll`.
+
+This file is [necessary](https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/) because GitHub Pages supports Jekyll ([jekyllrb.com](http://jekyllrb.com)), a static website builder, by default. You need the `.nojekyll` file because bookdown sites uses files or directories that start with underscores, and these files have a special status within the Jekyll framework. Because the **bookdown** HTML output is a static, standalone website, you need to tell GitHub that your website should *not* be built via Jekyll.
```
cderv, line 78: This is where the file needs to be inside the `/docs` folder. 

So it should be `file.create("docs/.nojekyll")` in your example. And `docs/` should not be cleaned, otherwise the file would need to be restored.

Other solution would be to create the file at build time, but this currently needs a hack because the file is not moved automatically to output dir. See https://github.com/rstudio/bookdown/issues/1190#issuecomment-871217855
 

This is why we need to provide a better way. 

I would advice your solution for now creating in the right directory. As it will be tracked by Git, a user would quickly see if this will is to be deleted during a cleaning. 

Note on cleaning: You need to clean you book output dir sometimes when working on a book because bookdown won't do that for you. There could be some conflicts sometimes. For example: You work on a chapter, you build, you commit, you are not happy and change the structure of the book which will lead to new html file name because new chapters. The old chapter does not exist anymore, but the HTML files still are. You will commit the new ones but it will not necessarily overwrite the old ones if the names are different. You will have old HTML files in your published book. For this a cleaning is required before rebuilding the book.

inst/examples/06-publishing.Rmd
```diff
@@ -19,42 +19,103 @@ bookdown::publish_book(render = 'local')
...
+
+* Delete the folders locally
+
+* From the terminal: 
+  * `git rm -r --cache _book/`
+  * `git rm -r --cache _bookdown_files/`
+  * `git commit -m 'Remove the rendered book files'`
+  * `git push origin main`
+
+:::
```
cderv, line 96: I think in this case you also want to delete them locally, at least `_book`, isn't it ? 

Your new book will be in `docs/`. 

But I guess it is easier for a user to do that itself than knowing how to remove from git.

cderv (commented):
Few comments below. I believe the part on `.nojekyll` is not correct. 

Also, on the whole chapter, your proposal is to separate Github Pages from Github Actions ? 

I have the same thought as for Netlify: I am wondering if we should have 
````
6. Publishing 
  |>> Github 
      |>> Github Pages
      |>> Github Actions
````

Now considering
* RStudio Connect
* Netlify
* GIthub Pages

maybe we need to replace the last part (now travis) to something like "Automate Publishing" and explain how to do it for the 3 service above. 

How would that be ? 

And as commented below, I would use `gh-pages` instead of `docs/` folder to automate the build and publish with GH pages. Am I the only one ?

inst/examples/06-publishing.Rmd
apreshill: I would definitely suggest the `gh-pages` route for folks who are comfortable with Make, gert, and CI systems like GHA (like yourself). I can draft that section, but could use a draft GHA workflow (much easier than Make/gert) since there is no `deploy_to_branch()` :)

inst/examples/06-publishing.Rmd
apreshill: Got it, I was thinking that was not going to work but then re-read the existing section (https://bookdown.org/yihui/bookdown/github.html) and thought I was wrong.

inst/examples/06-publishing.Rmd
apreshill: I was told by several on the tidyverse team that you had to do both (which is what this says)- so you mean only do step 1?

inst/examples/06-publishing.Rmd
cderv: Ok so we agree on this. 

I suggest we do the GHA part later - this would require to rework a bit a workflow, and I would like to show netlify and rstudio connect for all options. Best would be to have demo repo too with all this !  
So it is a bit of work to it right IMO. 

It is also possible to work on helper function to help deploy. We have already `publish_book()` for rstudio connect, we could have another one for GH Pages deployement (but it could also be directly in the Workflow template). 

I can work ont that... but later in another PR. 

How does that sounds ?

inst/examples/06-publishing.Rmd
cderv: No no what I am saying is that  `git rm -r --cache _book/` will not delete the `_book` folder. 
It will still be there but untracked. If you don't need it anymore (because you are using `docs/` now, it should be removed I think (`bookdown::clean_book(TRUE)`)

For `_bookdown_files`, I don't remember what it is used for. Doing `git rm -r --cache _bookdown_files/` will untrack eveything from Git, but the files will still be there on the system. 

I wasn't sure here if we are supposed to keep those or not. 

Anyway, if those command where proven useful in the past, then it is the right one to switch. Usually I would set up at the start of a project, so I don't really try to undo stuff. But just start fresh. 

Meaning: 
* `bookdown::clean_book(TRUE)` => start fresh bookdown project
* change config `output_dir` and other stuff
* `bookdown::render_book()` 
* Commit everything (new stuff and also result of first step deletion.)

The only time I am using `git rm -r --cache` if I am right is when I want to untrack something without deleting it in my worktree because I want to now ignore it in the future

yihui (commented):
Reading all these steps, I feel we should just provide an "official" Github action to automatically publish to either `docs/` or `gh-pages` (based on user's configuration). Then we won't need to document the complicated git commands or emphasize the importance of `.nojekyll` (or how to create/maintain it).

With that action, I think the Travis section can be deleted, since there won't be any obvious benefits with using Travis.

Perhaps we should wait till @cderv is back to work on this action, and then greatly simplify this section.

PR #1249, opened 2021-09-01
Put gitbook dependencies before extra_dependencies

### LINKED ISSUE #1101: extra_dependencies order in bookdown::gitbook
# Context

I am trying to make a custom output format that adds a few features to the default `gitbook` output (repository [here](https://github.com/InseeFrLab/utilitr-template)). The examples provided here all come from Github Actions build [from this repository](https://github.com/InseeFrLab/utilitr-template) 

Among the things I do, I add some `extra_dependencies` using some custom CSS and JS files (I have very limited knowledge in these two languages. I hope I will not state the obvious). This is [there](https://github.com/InseeFrLab/utilitr-template/blob/8b5021d455be870434c0956a90b48f7a0a9708ab/R/build_website.R#L56). However, when looking at the source code of the generated HTML (an example deployed with Netlify [here](https://awesome-volhard-d6d842.netlify.app/) ), I see that the order is 

```html
<link href="libs/utilitr-default-0.1.0/css/default.css" rel="stylesheet" />
<link href="libs/utilitr-default-0.1.0/css/style-utilitr.css" rel="stylesheet" />
<link href="libs/utilitr-default-0.1.0/css/icones-fa.css" rel="stylesheet" />
<link href="libs/utilitr-default-0.1.0/css/customize.css" rel="stylesheet" />
<script src="libs/utilitr-default-0.1.0/js/book.js"></script>
<script src="libs/jquery-2.2.3/jquery.min.js"></script>
<link href="libs/gitbook-2.6.7/css/style.css" rel="stylesheet" />
<link href="libs/gitbook-2.6.7/css/plugin-table.css" rel="stylesheet" />
<link href="libs/gitbook-2.6.7/css/plugin-bookdown.css" rel="stylesheet" />
<link href="libs/gitbook-2.6.7/css/plugin-highlight.css" rel="stylesheet" />
<link href="libs/gitbook-2.6.7/css/plugin-search.css" rel="stylesheet" />
<link href="libs/gitbook-2.6.7/css/plugin-fontsettings.css" rel="stylesheet" />
<link href="libs/gitbook-2.6.7/css/plugin-clipboard.css" rel="stylesheet" />
```

this means that some elements in our stylesheets might be overriden by elements defined in some stylesheets included in the `bookdown` package. I made an example [there](https://6050e968b9c06b063745988c--awesome-volhard-d6d842.netlify.app/) where I put some elements to define toc appearance in a CSS that is called in `extra_dependencies` ([code there](https://github.com/InseeFrLab/utilitr-template/commit/da1cfbd46e9af54076259be45b29ad75d1bda8a3)). Customized elements, e.g. text color in the TOC, are ignored (when looking at the output in Firefox).

The fix I found to force rendering is to put some elements in a `style.css` stylesheet called in the `_output.yml`. I used this fix to generate the aforementioned [webpage](https://awesome-volhard-d6d842.netlify.app/)  where I modify the appareance of the TOC.

In my opinion, using the `_output.yml` file is not ideal when developping a format within a package. This makes it complicated, for instance, to modify the behavior of the `gitbook` table of content within the package (one cannot assume that the user of the package always add a CSS in the  `_output.yml` YAML). 

# Why it matters ? (in my opinion, I am open to discussion)

I think the order of the dependencies matter because, when using a custom output format, one would expect the classes or properties defined in a few custom CSS to override some default `bookdown/gitbook` classes. Some attributes could still be set as `!important` but I am not sure this would always work. 

## A potential solution ?

I think a partial solution would be to change the order of  `extra_dependencies` in `bookdown::gitbook`  ([here](https://github.com/rstudio/bookdown/blob/2b7830f1e701df1c459f359c47c06cf3840e441e/R/gitbook.R#L30)) like this is done in `bookdown::bs4_book` ([here](https://github.com/rstudio/bookdown/blob/2b7830f1e701df1c459f359c47c06cf3840e441e/R/bs4_book.R#L70)):

```r
extra_dependencies = c(gitbook_dependency(table_css), extra_dependencies)
```

## Regarding scripts in `extra_dependencies`

However, I think this does not solve everything regarding JS. Some scripts are present at two locations in the generated HTML. When matching the following pattern, scripts are copied at the bottom of the HTML ([source here](https://github.com/rstudio/bookdown/blob/2b7830f1e701df1c459f359c47c06cf3840e441e/R/gitbook.R#L133 )):

```r
i = grep('^\\s*<script src=".+/gitbook([^/]+)?/js/[.a-z-]+[.]js"></script>\\s*$', head)
```

The scripts that I would define in `extra_dependencies` will not be duplicated (unless adopting a naming convention, which might be fine). This means you cannot override some functions if you want to change, let's say, the toolbar at the top. 

# Extra

Before posting the issue, I have followed the recommendations ;

- [x] [formatted your issue](https://yihui.org/issue/#please-format-your-issue-correctly) so it is easier for us to read?

- [x] included a minimal, self-contained, and reproducible example?

cderv:
so that extra_dependencies is not overridden by some Gitbook deps

This will close #1101

It should not be impactful and `extra_dependencies` must no be so used, and those who tried must have faced the issue. 

See discussion in #1101

Let's change the order.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,8 @@
 
 ## BUG FIXES
 
+- `extra_dependencies` in `gitbook()` will now be appended after Gitbook's dependencies so that it does not get overridden (thanks, @ThierryO, @linogaliana, #1101, #1248).
+
 - Fix an issue with Fenced Divs for Theorem & Proof environments in the Lua filter (thanks, @tchevri, #1233).
 
 - `gitbook(self_contained = TRUE)` was slow when the output contains base64-encoded images (thanks, @king2bob, #1236).
--- a/R/gitbook.R
+++ b/R/gitbook.R
@@ -31,7 +31,7 @@ gitbook = function(
   gb_config = config
   html_document2 = function(..., extra_dependencies = list()) {
     rmarkdown::html_document(
-      ..., extra_dependencies = c(extra_dependencies, gitbook_dependency(table_css, gb_config))
+      ..., extra_dependencies = c(gitbook_dependency(table_css, gb_config), extra_dependencies)
     )
   }
   if (identical(template, 'default')) {


yihui (approved):
Thanks!

PR #1253, opened 2021-09-06
allow footnotes to note be inline in `bs4_book()`

Pindar777:
the new switch `fn_gitbook` makes it possible to disbale `tweak_footnotes` which results in gitbook-style footnotes.
Perhaps some additional tweaking is necessary to grant that the footnote-div is inserted always right before the cahpter-nav-div.
Default-value is `FALSE`


cderv:
Hi, 

What is the issue exactly you are trying to solve here ? As there is no issue open to discuss the feature, I may have missed something. 

`bs4_book` was designed with making footnote style citation inline, but you can already not use a footnote style and have citation at the end of the book only, or since #1185 at each chapter by setting `split_bib: TRUE` (unless there is an issue with this and it does not work anymore.)
https://bookdown.org/yihui/bookdown/html.html#referencesbibliography

Please can you describe what you are trying to solve ?

Pindar777:
Hi @cderv,

my pull request is about introducing a new feature and by this also solving the issue of hidden footnotes in case when he execution of JS is not allowed on the server.

`tweak_footnotes` puts each footnote inside a container which pops up when clicking on the number of the footnote.
My switch in contrast leaves the footnotes at the end of each page. Clicking the footnote number moves the cursor to the footnote-entry and there is a return arrow to get back to the main text.

Hence `gitbook-style-footnotes` are working without JS and the text of the footnotes is searchable inside the browser window. Both features could come in handy and are achives by adding `fn_gitbook: TRUE` in the _output.yml.

Cheers

cderv:
> tweak_footnotes puts each footnote inside a container which pops up when clicking on the number of the footnote.

This will indeed happen if you use a footnote style CSL like a `chicago-fullnote-bibliography.csl`. It should not otherwise. 

> My switch in contrast leaves the footnotes at the end of each page. Clicking the footnote number moves the cursor to the footnote-entry and there is a return arrow to get back to the main text.

This seems like what `gitbook()` is doing, and if I understand correctly, you should already be able to achieve this by with last version of **bookdown** as `split_bib` is now configurable: 

* Using a CSL that it not footnote style like you would do in `gitbook()`
* Using `split_bib: true` like in `gitbook()`
* Using `link-citations: yes` in YAML of `index.Rmd` like with `gitbook()`

Is that not working for you ? 

What is missing ?

Pindar777:
@cderv Thank you for the proposal of a parameter set.
I already had `link-citation: yes`in ìndex.Rmd`, no CLS (?) and `split_bib: true` in `_output.yml`.
I'm using the gitbook-index.Rmd-parameter `biblio-style: apalike`. However, this should only alter the reference section, or?

When looking in the code of `bs4_cook.R` and `tweak_footnotes` I cannot find a condition that changes with csl:

```
container <- xml2::xml_find_all(html, ".//div[@class='footnotes']")
  if (length(container) != 1) {
    return()
  }

  # Find id and contents
  footnotes <- xml2::xml_find_all(container, ".//li")
  id <- xml2::xml_attr(footnotes, "id")
  xml2::xml_remove(xml2::xml_find_all(footnotes, "//a[@class='footnote-back']"))
  contents <- vapply(footnotes, FUN.VALUE = character(1), function(x) {
    paste0(as.character(xml2::xml_children(x)), collapse = "")
  })

  # Add popover attributes to links
  for (i in seq_along(id)) {
    links <- xml2::xml_find_all(html, paste0(".//a[@href='#", id[[i]], "']"))
    xml2::xml_attr(links, "href") <- NULL
    xml2::xml_attr(links, "id") <- NULL
    xml2::xml_attr(links, "tabindex") <- "0"
    xml2::xml_attr(links, "data-toggle") <- "popover"
    xml2::xml_attr(links, "data-content") <- contents[[i]]
  }

  # Delete container
  xml2::xml_remove(container)
```

That's why I introduced the new parameter and disable the tweaking. By this only I could achieve the gitbook-behavior.

Pindar777:
The switch could also be introducec inside of `bs4_chapter_tweak()`. But there no csl-condition, too.

cderv:
Here is a reproducible example of what I am describing. 
* Take the new bs4_book template
* Render it to see how it works with default configuration which use inline footnote
* Then, remove CSL configuration, add the `split_bib` and `link_citation`. Render again and see the result. 
* If you remove the CSL only, you will have a default style for citation and references only at the end of the book. 

````r
dir.create(tmp_dir <- tempfile())
book <- bookdown::create_bs4_book(tmp_dir)
owd <- setwd(book)

# Render the default template which uses citation in footnote, 
# which are placed inlined by bs4_book
res <- bookdown::render_book(tmp_dir, bookdown::bs4_book())
# Preview the book
servr::httd(dirname(res))

# Look at chapter 5 and citations will be inlined.

# Clean the book
bookdown::clean_book(TRUE)

# Remove CSL file in YAML and add link-citations
xfun::gsub_file("index.Rmd", pattern = "^csl:.*$", replacement = "link-citations: yes")

# Render the book using split_bib = TRUE
res <- bookdown::render_book(tmp_dir, bookdown::bs4_book(split_bib = TRUE))

# Preview the book
servr::httd(dirname(res))

# Look at chapter 5 again: Citation will be a link toward the end of the page.


# Cleaning the reprex
setwd(owd)
unlink(tmp_dir)
````

Can you try this and see how it differs from your try ? 
Can you confirm this works for you ? 

Regarding how the tweak function is working, it will only apply when note citation style are used. The style of citation is inside the CSL file. 
You can try to add chicago fullnote style to `gitbook()` output and see what it will look like: 
````r
dir.create(tmp_dir <- tempfile())
book <- bookdown::create_gitbook(tmp_dir)
owd <- setwd(book)

# Add CSL style to use chicago-fullnote
xfun::gsub_file("index.Rmd", 
                pattern = "^---$", 
                replacement = "csl: https://www.zotero.org/styles/chicago-fullnote-bibliography\n---")

# Render gitbook
res <- bookdown::render_book(tmp_dir, bookdown::gitbook())

# Preview the book
servr::httd(dirname(res))

# Look at chapter 5 again: Citation will be like any footnote, hence footnote style citation.

# Cleaning the reprex
setwd(owd)
unlink(tmp_dir)
````

Looking at this example, you'll see that `chicago-fullnote` CSL will make citation appear like any footnotes at the end of each chapter. This is what `tweak_footnotes` is looking for. It won't find that with a non-footnote citation style. 

I hope this helps understand what I am describing in this thread.

**Comment by Pindar777** [comment]
Hi @cderv,

your reprex and descriptions helped to understand the overall setting. Thank you :-)

Please find attached my output of **three** variants where the first and the second correspond to your reprex and the last one uses my switch for the **standard footnote**:

` res <- bookdown::render_book(tmp_dir, bookdown::bs4_book())`
bs4_book_default_dummy

`res <- bookdown::render_book(tmp_dir, bookdown::bs4_book(split_bib = TRUE))`
bs4_book_split_bit_link_citation_no_csl_dummy

**and**

`res <- bookdown::render_book(tmp_dir, bookdown::bs4_book(split_bib = TRUE, fn_gitbook = TRUE))`

bs4_book_split_bit_link_citation_no_csl_AND_fn_gitbook_dummy

**Comment by cderv** [comment]
Oh so know I understand. Thanks for the screenshot. You are not talking about footnotes for citations but usual footnotes ! Sorry I missed that. 

Yes, `bs4_book()` is an HTML format designed to have footnotes inlined instead of having them at the end of the page. You are proposing a switch to opt out this mechanism. 

Why exactly to we want to change this ? I fear that by starting to do so, we will end up with having a lot of argument or options to change default behavior of this format. `bs4_book()` is an **opinionated** format based Bootstrap. You are mentioning this would solve no JS allowed but Bootstrap needs JS and `bs4_book()` will require JS I believe. Is the popover for footnotes the only thing that requires JS ? I don't think we'll want to aim at providing a `bs4_book()` format JS-free. So if this is the aim, I would prefer not to start.

Anyway, I'll have a closer look on this now that I understand. Just so you know I don't really like the name `fn_gitbook` here, this would probably changed. And please, can you document so that the CI check passes ? 

Thank you!

**Comment by Pindar777** [comment]
@cderv Great :-) 
Yes, it's a working title so to speak...

The starting point was that "the click" did not work. However, in the meantime I really prefer the classical footnote style in order to have standard browser search access to all content. The decision clearly depends on the type of bs4_book you are working on. Having the choice would be great.

What do I have to do to solve the tests and change the documentation?

**Comment by Pindar777** [comment]
@cderv Documentation is matching now

**Review by cderv — APPROVED** [review]

**Comment by cderv** [comment]
Thanks for the PR. I have changed the name to `footnotes_inline` and set the default to be `TRUE` (current behavior). 

You can now do 
```yaml
bookdown::bs4_book:
    footnotes_inline: false
```

to keep the footnotes at the bottom.

PR #1277, opened 2021-10-26
Add CSL styling to bibliography in `bs4_book()`

Selbosh:
The default CSS styling for the bibliography / reference list is very difficult to read, because there is no spacing between entries and it does not do anything for the `.hanging-indent` class. This short PR makes them a bit more readable. If anyone doesn't like this styling then they can obviously add their own styles later, but I find it hard to imagine that most people would be satisfied with the current default look.


pat-s:
One of ours books is also affected by this - would be great to see an upstream fix :) 

For the time being, using

```css
div.csl-entry {
  clear: both;
  margin-bottom: 10px;
}
``` 

makes references somewhat readable.

inst/templates/bs4_book.html
```diff
@@ -34,6 +34,33 @@
...
+  <style type="text/css">
+    /* for pandoc --citeproc since 2.11 */
+    div.csl-bib-body { }
+    div.csl-entry {
+      clear: both;
+    $if(csl-entry-spacing)$
+      margin-bottom: $csl-entry-spacing$;
+    $endif$
+    }
+    .hanging div.csl-entry {
```
salim-b, line 47: This should be `.hanging-indent` instead of `.hanging` (at least with latest Pandoc 2.17.1.1).

cderv: Thanks I'll update this.

cderv: Latest Pandoc is having the same as here though 🤔 
https://github.com/jgm/pandoc/blob/581c94913fa055fa0d925db893991f01647b37f6/data/templates/styles.html#L187-L190

Hanging indent is defined elsewhere in the template 
https://github.com/jgm/pandoc/blob/581c94913fa055fa0d925db893991f01647b37f6/data/templates/styles.html#L168

We could add this too though.

cderv: And I forgot this was also discussed previously above https://github.com/rstudio/bookdown/pull/1277#discussion_r737289032

cderv (approved):

cderv:
Thanks @Selbosh for the PR 

Sorry for the delay on this everyone.

@pat-s we've put the same as in Pandoc's tempalte. `margin-bottom` is set by Pandoc depending on the entry in the CSL file for the option `entry-spacing`. Example with https://www.zotero.org/styles/stavebni-obzor 
```yaml
csl: https://www.zotero.org/styles/stavebni-obzor
````

If your CSL doesn't have that and you want it, you need to add the CSS rule (or set `csl-entry-spacing` pandoc's variable in `index;Rmd` - it won't probably be overriden). 

Thanks everyone !

**Comment by pat-s** [comment]
@cderv Thanks!

In our case we use the oldschool `biblio-style: apalike` in `index.Rmd`. I see that [apalike.csl](https://github.com/robertour/apalike-zotero/blob/master/apalike.csl) does not have a `entry-spacing` definition. 

I've also tried setting `entry-spacing` in `index.Rmd` but that did not help. I have probably done something wrong but the CSS class definition workaround seems to do the job for now :)

**Comment by cderv** [comment]
I think this would work 
````yaml
csl: https://raw.githubusercontent.com/robertour/apalike-zotero/master/apalike.csl
csl-entry-spacing: 2em
````

But adding your own CSS is as good as this, moreover because `csl-entry-spacing` is an undocumented option.

**Comment by pat-s** [comment]
Thanks! Just to understand the details a bit better: Do you know that the difference is between

```
csl: https://raw.githubusercontent.com/robertour/apalike-zotero/master/apalike.csl
```

and 

```
biblio-style: apalike
```

for `rmarkdown`/pandoc?

**Comment by cderv** [comment]
`csl` is used with Pandoc citeproc:  for almost all formats: https://pandoc.org/MANUAL.html#specifying-a-citation-style 
It could be a url or a local file (`apalike.csl` here)

`blibio-style` is used by natbib or biblatex, for LaTeX/PDF output : https://pandoc.org/MANUAL.html#biblatex-bibliographies
`biblio-style: apalike` means use `apalike.bst` file.

Does it help ?

File: inst/templates/bs4_book.html (hunk 1 above)
Comment (salim-b) [reply inst/templates/bs4_book.html:51]: Ok, thanks for the hints.

Just for the record, this is the relevant bibliography div generated by bookdown (latest master):

```html
<div id="bibliography" class="section level1 unnumbered">
  <h1>Bibliography<a class="anchor" aria-label="anchor" href="#bibliography"><i class="fas fa-link"></i></a>
  </h1>
  <div id="refs" class="references csl-bib-body hanging-indent">
    <div id="ref-citationkey1" class="csl-entry">
      ...
    </div>
    <div id="ref-citationkey2" class="csl-entry">
      ...
    </div>
    <div id="ref-citationkey3" class="csl-entry">
      ...
    </div>
  </div>
</div>
```

None of the elements has the `.hanging` class as you can see.

cderv: Yes I know - Bibliography part is generated by Pandoc and this how Pandoc does thing. So maybe there is a case (I don't know which) where this is required. With specific CSL attributes maybe ? IDK. 
 
You could open an issue there if you thing this is incorrect or a duplicate as there is also `div.hanging-indent`. Or ask in pandoc discuss. 

IMO it is better for **rmarkdown** to not change Pandoc on this, especially if putting the same as in template is working.

PR #1278, opened 2021-10-29
Add option to prevent search engine indexing

muskuloes:
Hello,

For the use case where one is working on an unfinished online book, I think the option to prevent indexing from search engines would be nice. This can be done with the [`robots` meta tag](https://developers.google.com/search/docs/advanced/crawling/block-indexing). I added the tag option to the gitbook and bs4_book HTML templates. 
A list of the tag values supported by various search engines can be found here: https://yoast.com/robots-meta-tags/

### DIFF
--- a/inst/templates/bs4_book.html
+++ b/inst/templates/bs4_book.html
@@ -23,6 +23,7 @@
   $if(twitter-handle)$<meta name="twitter:site" content="@$twitter-handle$" />$endif$
   $if(description)$<meta name="twitter:description" content="$description$" />$endif$
   $if(cover-image)$<meta name="twitter:image" content="$url$/$cover-image$" />$endif$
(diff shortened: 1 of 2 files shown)


yihui (commented):
I tend not to add more top-level YAML options unless there are no workarounds, because it would need documentation and we also need to make sure that Pandoc won't use this option for a different purpose in the future.

In this case, I think you can use the `includes` option of `gitbook` or `bs_book`, e.g., create a `robots.html`:

```html
<meta name="robots" content="noindex" />
```

Then include it:

```yaml
bookdown::gitbook:
  includes:
    in_header: robots.html
```

Alternatively, you can probably also use `robots.txt`, which may be a little simpler.

Thanks!

muskuloes:
Alright, I'd have the PR closed then.

PR #1319, opened 2022-03-17
Tweaking fuse.js configuration to improve search results

cderv:
Context: https://community.rstudio.com/t/bookdown-search/131042/12

* Add `ignoreFieldNorm: true` to fuse js


jtbayly:
Hey @cderv ,

I finally got around to testing this in the javascript console. I didn't specifically test this version of the code, since it was easy to just set ignoreFieldNorm: true in the browser. 

It solved the problem we were discussing here: https://community.rstudio.com/t/bookdown-search/131042

If you were to make this change, it would mean that shorter fields wouldn't be scored higher than longer fields. Given that... 

1. we are currently not returning all of the exact matches, which seems to me to be a clear failure, meaning this change will cause some search results to be included that *are* high quality, and
2. my limited testing doesn't show returning low quality results after this

 I think it makes sense to make this change at least in the dev version so it can receive wider testing.

jtbayly:
I should have mentioned, we'd want to do this in the gitbook version, too, right? I'm not using the bs4_book theme.

cderv:
Thanks ! 

> we'd want to do this in the gitbook version, too, right? 

`gitbook()` is using default configuration for fuse.js. That is to the user to define the config desired in YAML. `bs4_book()` is more opinionated so hence this change there.

> which seems to me to be a clear failure, meaning this change will cause some search results to be included that are high quality,

Can we improve this ?

jtbayly:
Sorry that sentence was unclear. What I meant is that in my limited testing I was only able to find improvements (exact matches from longer paragraphs showing up in search results as expected).

PR #1321, opened 2022-03-22
Fix/add links to webshot documentation

salim-b:
There doesn't seem to be a vignette anymore, only a pkgdown article.


cderv:
Oh indeed it seems that it has been not there from some time now https://github.com/wch/webshot/commit/2eb6dafd4f93df36527a4c20dc828d0ebfbe89cf

thanks for catching that!

PR #1324, opened 2022-04-04
resolve format in `render_book()` directly for `rmd_files` config resolution

### LINKED ISSUE #1323: Book in pdf format with rmd_files: in _bookdown.yml fails to compile
Hi,
I developed a {bookdown} template for theses at a certain institution, and some user now reported the following error when trying to compile the book via the "Knit" button in RStudio:

```r
Error in files2[[format]] : 
  attempt to select less than one element in get1index
Calls: <Anonymous> ... <Anonymous> -> <Anonymous> -> render_book -> source_files
Execution halted
```

However, using the "Build Book" button from the "Build" tab in RStudio seemed to work fine.
Interestingly, I noticed that this is likely due to using `rmd_files: [...]` in the `_bookdown.yml` file, as if I removed that, it compiles without issue either way.

You can find the repository at that specific point in time [here](https://github.com/ellessenne/uol-thesis/tree/fa7b3913b95ce4fa69ea2c818e8d277fff016917), in the `bookdown` folder (I now removed the `rmd_files` part).

I tried installing the latest version from GitHub, but the issue still persists.

I hope this helps, and thanks for the fantastic work on {bookdown}!

Alessandro

cderv:
This fixes #1323 

Currently, `render_book()` is called with `output_format = NULL` by default. This will let `rmarkdown::render()` do the resolution to the first format as with other docs. 

However, `rmd_files` configuration in `_bookdown.yaml` can require the `format` to resolve the files to pass to `render()`. 
This is really the case when a per-format `rmd_files` config is provided - quite rare probably as this is the first time this issue is encountered. 

Build book button is working fine as the format is resolved by the IDE (not sure with which function) and `output_format` name is added in the call to `render_book()`. 

First idea here is to fully resolve the format (using `resolve_output_format()`), when none is provided - so basically doing what `rmarkdown::render()` will do later.  This require explicitly passing `output_options` argument.

Other approach would be to use `rmarkdown::all_output_formats()` like `output_format = "all"` but keep only the first one. I am just not sure that is equivalent to what `render()` is doing.  I don't want to break current `render_book()` calls to fix this specific issue. 

Does current solution looks ok to you ?

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -14,7 +14,7 @@
 #'   project directory. For filenames, if \code{preview = TRUE}, only files
 #'   specified in this argument are rendered, otherwise all R Markdown files
 #'   specified by the book are rendered.
-#' @param output_format,...,clean,envir Arguments to be passed to
+#' @param output_format,output_options,...,clean,envir Arguments to be passed to
 #'   \code{rmarkdown::\link{render}()}. For \code{preview_chapter()}, \code{...}
 #'   is passed to \code{render_book()}. See \code{rmarkdown::\link{render}()}
 #'   and \href{https://bookdown.org/yihui/bookdown/build-the-book.html}{the
@@ -56,7 +56,7 @@
 #' bookdown::render_book("my_book_project")
 #' }
 render_book = function(
-  input = ".", output_format = NULL, ..., clean = TRUE, envir = parent.frame(),
+  input = ".", output_format = NULL, output_options = NULL, ..., clean = TRUE, envir = parent.frame(),
   clean_envir = !interactive(), output_dir = NULL, new_session = NA,
   preview = FALSE, config_file = '_bookdown.yml'
 ) {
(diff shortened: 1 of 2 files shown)


cderv:
After all adding `output_options` as argument implies more changes. 

Maybe my second solution is better

````diff
diff --git a/R/render.R b/R/render.R
index ac87d259..46c01ab6 100644
--- a/R/render.R
+++ b/R/render.R
@@ -77,9 +77,11 @@ render_book = function(
   if (is.list(output_format)) {
     format = output_format$bookdown_output_format
     if (!is.character(format) || !(format %in% c('latex', 'html'))) format = NULL
-  } else if (is.character(output_format)) {
-    if (identical(output_format, 'all')) {
+  } else if (is.null(output_format) || is.character(output_format)) {
+    if (is.null(output_format) || identical(output_format, 'all')) {
       output_format = rmarkdown::all_output_formats(input)
+      # when no format provided, return name of the first resolved
+      if (is.null(output_format)) output_format = output_format[[1]]
     }
     if (length(output_format) > 1) return(unlist(lapply(output_format, function(fmt)
       xfun::Rscript_call(render_book, list(
````

if `rmarkdown::all_output_formats()[[1]]` correctly returns the format that `rmarkdown::render()` with no output format would have chosen. It is not that obvious from the code base as they do share a single function for the logic, but reading both seems to confirm this ok. 

Maybe you are seeing another fix after all to the original issue ?

yihui (approved):
Looks good!

PR #1329, opened 2022-04-06
Fix TOC issues with pandoc 2.18 in gitbook()

### LINKED ISSUE #1326: New and disturbing span after part
Hi!

In the book I use the standard PART-s injection: `# (PART) MYPART {-}`

Having updated somewhere today, I saw in the book these span, which did not exist before.

I can't figure out where they came from. They remain even if I remove my css or go back to previous versions of bookdown.

Please fix/tell me how to remove this span

Thanks!

image

### LINKED ISSUE #1327: PART is broken with Pandoc 2.18
This comes for the change in the TOC id in last Pandoc version.

cderv:
Fixes #1327 and close #1326 by dealing with empty spans introduced in gitbook. 

basically for #1327 we had our usual regex issue following Pandoc 2.187 changes. There was the PART issue, but also the unnumbered chapter not being dealt with correctly. 

We did not detect sooner because we did not have tests for this. So I took the time to add some more test for this - took me a while to master more XPATH but that will be useful for further tests.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.25.4
+Version: 0.25.5
 Authors@R: c(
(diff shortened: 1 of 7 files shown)


yihui (approved):
Great! Thanks a lot!

PR #1333, opened 2022-05-02
Add syntax highlighting for imports and built-ins tokens

briandk:
## TL;DR

Currently bookdown's `bs4_book()` output format doesn't apply any particular syntax highlighting to code tokens of class "import" and "built-in". This PR tries to remedy that by applying sensible styles to tokens of those classes.

## Background

Much of the background on this issue can be found at https://github.com/jgm/skylighting/issues/147, but in short:

Right now the imports in a Python code chunk don't have any visually distinguishing features. A statement like this:

```python
from plotnine import geom_text
```

appears like this, with no coloring or font weight changes:

```
from plotnine import geom_text
```

The reason this code doesn't get styled by `bs4_book.css` is that the CSS file doesn't actually *have* any extra styling information for those types of tokens (import tokens, specifically, but the same is true for tokens of class built-in). Note the empty ruleset for `code span.im` here:

https://github.com/rstudio/bookdown/blob/6adacc3363315a9a49eb4fac370125a3d37ecaa5/inst/resources/bs4_book/bs4_book.css#L450

## What I've tried to do in this PR

I tried my best to preserve the current aesthetic of bs4_book's syntax highlighting palette while adding support for additional token types.

I took my inspiration from [Pygments defaults][pygments], trying to match colors and font weights without introducing any new colors to bs4_book's palette.

[pygments]: https://pygments.org/demo/?lexer=python&style=default&formatter=html&code=from+typing+import+Iterator%0A%0A%23+This+is+an+example%0Aclass+Math%3A%0A++++%40staticmethod%0A++++def+fib%28n%3A+int%29+-%3E+Iterator%5Bint%5D%3A%0A++++++++%22%22%22+Fibonacci+series+up+to+n+%22%22%22%0A++++++++a%2C+b+%3D+0%2C+1%0A++++++++while+a+%3C+n%3A%0A++++++++++++yield+a%0A++++++++++++a%2C+b+%3D+b%2C+a+%2B+b%0A%0Aresult+%3D+sum%28Math.fib%2842%29%29%0Aprint%28%22The+answer+is+%7B%7D%22.format%28result%29%29

### DIFF
--- a/inst/resources/bs4_book/bs4_book.css
+++ b/inst/resources/bs4_book/bs4_book.css
@@ -440,14 +440,14 @@ code span.er {color:#ff0000;}
 code span.bn {color:#a1024a}
 code span.al {color:#ff0000;}
 code span.va {color:#19177c}
-code span.bu {}
+code span.bu {color: #007faa;}
 code span.ex {}
 code span.pp {color:#bc7a00}
 code span.in {color:#545454;}
 code span.vs {color:#008000}
 code span.wa {color:#545454; font-style: italic}
 code span.do {color:#ba2121; font-style: italic}
-code span.im {}
+code span.im {color:#007faa; font-style: bold;}
 code span.ch {color:#008000}
 code span.dt {color:#aa5d00}
 code span.fl {color:#a1024a}
@@ -456,7 +456,7 @@ code span.cv {color:#545454; font-style: italic}
 code span.cn {color:#d91e18}
 code span.sc {color:#008000}
 code span.dv {color:#a1024a}
-code span.kw {color:#007faa;}
+code span.kw {color:#007faa; font-style: bold}
 
 /* Misc typography ---------------------------------------------- */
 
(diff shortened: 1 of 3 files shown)


inst/resources/bs4_book/bs4_book.css
```diff
@@ -456,7 +456,7 @@ code span.cv {color:#545454; font-style: italic}
 code span.cn {color:#d91e18}
 code span.sc {color:#008000}
 code span.dv {color:#a1024a}
-code span.kw {color:#007faa;}
+code span.kw {color:#007faa; font-weight: bold;}
```
cderv, line 459: This will impact default look on current bs4_book() as `kw` style is used with R already. 

Is this important to have bold ?

cderv (commented):
Thanks. 

`bu` and `im` are not used for highlighting R code using **downlit** in `bs4_book`. So I don't mind changing them. 

For `kw`, it would have an impact. I would leave it at is.

inst/resources/bs4_book/bs4_book.css
briandk: @cderv Bolding would bring the syntax highlighting style more in line with what Pygments does, but I don't think it's necessary; just consistent with a particular style.

I removed the bolding in a7832dac3bf5586117026514ff33c90f1100eb8a

cderv (approved):
Thanks ! I'll add NEWS bullet and merge

PR #1346, opened 2022-06-14
Support 'split_by' as numeric

lcougnaud:
This Pull Request includes the extension of the 'split_by' parameter, to support splitting for section level higher than 'section' (issue #1347).
The 'split_by' parameter can be specified as [X] (or [X]+number) such as the report is split into separate html files by level [X] of sectioning (without/with numbering)
For example, the report can be split for a subsection, i.e. a header 1.1.1 if `split_level = 3`.


katrinabrock:
I would also be interested in this! Might rebase your PR and test it out.

PR #1351, opened 2022-06-22
Relax `index.Rmd` constraint and correctly support edge case

### LINKED ISSUE #1349: render_book("filename") complains about missing "index.Rmd"
I have a project in a separate folder that contains three files (and only these three files):

_foo.rmd`_

    # Chapter Foo

_bar.rmd_

    # Chapter bar

and __bookdown.yml_:

    book_filename: foobar
    
    rmd_files: ["foo.rmd", "bar.rmd"]

I attempt to knit the book as

```bash
Rscript -e "bookdown::render_book('foo.rmd')"
```
This gives me an error
```
Error in stop_if_not_exists("index.Rmd") : 
  Some files were not found: index.Rmd
Calls: <Anonymous> -> stop_if_not_exists
Execution halted
```
So bookdown is looking for _index.Rmd_ although this file is nowhere in my project, and `?render_book` states that

> ...
> input: A directory, an input filename or multiple filenames. For a
>     directory, ‘index.Rmd’ will be used if it exists in this
>     (book) project directory. For filenames, if ‘preview = TRUE’,
>     only files specified in this argument are rendered, otherwise
>     all R Markdown files specified by the book are rendered.

This occurs in R 4.2, bookdown 0.27 and also the github version 0.27.1.  It worked in R 4.1, bookdown 0.24.

For a larger project a workaround seems to be to create an empty _index.Rmd_ file, but not for this simple example.

```
> xfun::session_info()
R version 4.2.0 (2022-04-22)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.4 LTS

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
  LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
  compiler_4.2.0  graphics_4.2.0  grDevices_4.2.0 parallel_4.2.0  stats_4.2.0     tools_4.2.0     utils_4.2.0     xfun_0.31      
```

cderv:
This should close #1349 

This PR tries to achieve two things: 

* Supports `index.Rmd` and `index.rmd` as main file expected in a bookdown project
* Still render a book even if these file are not provided (basically `render_book("my_file.Rmd")` where `my_file.Rmd` plays the role of index file but with another name. 

I am opening this PR as a draft to get some first feedback @yihui. I also need to check that report in community (https://community.rstudio.com/t/render-book-error-stop-if-not-exists-and-some-files-were-not-found/139927/19) is taken into account.


R/bs4_book.R
```diff
@@ -679,8 +679,10 @@ check_packages <- function(pkgs) {
 preview_book <- function(path = ".", output = "bookdown::bs4_book") {
   old <- setwd(path)
   on.exit(setwd(old))
-
-  render_book("index.Rmd",
+  if (is_empty(index <- get_index_file())) {
+    stop(sprintf('`index.Rmd` or `index.rmd` was not found in path = "%s".', path))
```
yihui, line 683: I see this error message "index.Rmd or index.rmd" has appeared multiple times in this PR, and I'd prefer that we just use "index.Rmd` in the message. In other words, we don't emphasize "index.rmd". This is because I think the vast majority of users would probably use "index.Rmd", and it's not worth their brain memory to remember that "index.rmd" is also supported.

R/utils.R
```diff
@@ -386,17 +386,47 @@ serve_book = function(
...
+    stop("`serve_book()` expect `index.Rmd` or `index.rmd` in the book project.", call. = FALSE)
+  }
+  rebuild(index, preview_ = FALSE)  # build the whole book initially
   servr::httw('.', ..., site.dir = output_dir, handler = rebuild)
 }
 
+index_files <- xfun::with_ext("index", c("Rmd", "rmd"))
+
+get_index_file <- function() {
+  xfun::existing_files(index_files, first = TRUE, error = FALSE)
```
yihui, line 400: To be truly case-insensitive, I recommend that we use `list.files('.', '^index[.]Rmd$', ignore.case = TRUE)` here. That way, filenames like `index.RMD` will also be supported.

yihui (approved):
Just two small suggestions. Thanks!

R/utils.R
cderv: I did the change. We will just only select the first in alphabetical if there are several index files. 

But one is expected only, so I added a warning really.

cderv: It also required to check dynamically for index each time we compare for preview
https://github.com/rstudio/bookdown/pull/1351/files#diff-0d68ae244e4a8c5ddc391c48f5399a91f18bc2a955f098e0e66069c607b444f4R213

Should I cache the values of index files now ? 

This was why previously I used `index_files <- xfun::with_ext("index", c("Rmd", "rmd"))`

yihui: I think `list.files()` should be fast enough, so may not be worth caching. It's up to you.

`list.files(recursive = TRUE)` could be slow, though.

PR #1125, opened 2021-04-04
Allowing sharing on whatsapp

prdm0:
I find it very useful to share the link to the url that hosts the book via whatsapp. I am sure that this feature will appeal to many people.

PR #1353, opened 2022-06-25
Use actual Font Awesome 4.7.0 CSS

snipfoo:
Hello,

Commit 4ebf32a from #984 changed the version number of Font Awesome in [`inst/resources/gitbook/css/style.css`](https://github.com/rstudio/bookdown/blob/main/inst/resources/gitbook/css/style.css), but did not update the CSS with the actual contents of the Font Awesome 4.7.0 CSS. Consequently, icons from Font Awesome 4.7.0 (such as, e.g., `fa-usb`) were still undefined in the CSS, even though they were in the TTF font file.

This PR proposes to update `style.css` with all the definitions of Font Awesome 4.7.0.

I couldn't find any non-minified source file for `style.css`, so I had to edit the contents of `css/font-awesome.min.css` from the official 4.7.0 release ([font-awesome-4.7.0.zip](https://fontawesome.com/v4/assets/font-awesome-4.7.0.zip)) directly into the minified `style.css`:

- I first removed the rules from the Font Awesome 4.1.0 CSS. This was slightly tricky, as there were bits of non-Font Awesome rules in between Font-Awesome-specific rules:
  - on [line 9](https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/inst/resources/gitbook/css/style.css#L9), columns 257-351, there were rules for `.book .book-header` and `.book .book-summary`;
  - on [line 9](https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/inst/resources/gitbook/css/style.css#L9) as well, starting from column 20285 until the end of the line (col. 35783), there were rules for `.book-langs-index` and other things.

  All these extra rules are now on [line 10](https://github.com/rstudio/bookdown/blob/a8e53deec0c16870a0cae2de841e293d67bfd809/inst/resources/gitbook/css/style.css#L10), to keep them separated from the Font Awesome rules.

- Some rules of the Font Awesome 4.1.0 CSS had also been merged with other rules (such as the `.fa{-webkit-font-smoothing:antialiased}` and `.fa-ul>li{position:relative}` rules on [line 1](https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/inst/resources/gitbook/css/style.css#L1) or the `.fa,.fa-stack{display:inline-block}` rules on [line 6](https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/inst/resources/gitbook/css/style.css#L6)). I removed those as well (after making sure they came from FA 4.1.0 and that they still existed in FA 4.7.0), so as to keep all the Font Awesome rules in the same place.

- I then replaced [lines 7-9](https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/inst/resources/gitbook/css/style.css#L7-L9) with lines 2-4 from `font-awesome.min.css` version 4.7.0.

- I updated the path to the font file in the `@font-face` declaration, keeping only the source declaration for the TTF file (as it is the only one provided by the `gitbook` style).

- Finally, on [lines 13-15](https://github.com/rstudio/bookdown/blob/7e9b46fee109f06f86962d6a78b801d3491c94aa/inst/resources/gitbook/css/style.css#L13-L15), I removed the rule for the `fa-whatsapp` icon introduced by #1125, since this icon is now properly defined in Font Awesome 4.7.0 and doesn't it an extra rule (see [line 9](https://github.com/rstudio/bookdown/blob/a8e53deec0c16870a0cae2de841e293d67bfd809/inst/resources/gitbook/css/style.css#L9) in the updated file, columns 23163-23198).

Sorry for this lengthy and tedious description, but commit diffs on minified files are not really self-explicit :/

In any case, thank you very much for this useful tool!

snipfoo


yihui (approved):
Sounds good to me. I'll let @cderv have another look. Thank you very much!

cderv (approved):
Thanks a lot! That is great. 

Updating fontawesome inside gitbook format is a bit tricky, as I believe it was provided by the Gitbook format, which is not more maintained and available. we need to do the updating ourselves

> there were bits of non-Font Awesome rules in between Font-Awesome-specific rules:

I believe those part where from the Gitbook format, thant `bookdown::gitbook()` is based off. 

Ideally, I would like the format to use **fontawesome** package now to get the correct updates of the icon, but we need to make the specific book CSS on its own and check it working fine. 

Anyway, this is for another day. Thanks a lot for this work !

snipfoo:
Great! Thank you very much to both of you for the kind feedback and for the merge! :)

snipfoo

Issue #1354, opened 2022-06-29
[FR] Update MathJax src in gitbook template to latest version of 2.7


scarnecchia:
PR #937 is looking into making the MathJax version customizable. In the meantime, the template files reference the version hosted at https://mathjax.rstudio.com/latest/. Currently this is MathJax.js 2.7.2. Versions prior to 2.7.4 contain an XSS vulnerability ([CVE-2018-1999024](https://nvd.nist.gov/vuln/detail/CVE-2018-1999024))

There's also a issue open to patch the embedded version in [RStudio](https://github.com/rstudio/rstudio/issues/11535).

A simple way to patch this issue would be to update the src in the gitbook (and bs4) templates to point to another source to ensure that the resulting webpages are pulling the latest version of the 2.7.x family, such as: https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.9/MathJax.js?config=TeX-MML-AM_CHTML

Other than the fact that bookdown is maintained by RStudio, is there a particular reason for pointing to https://mathjax.rstudio.com/latest/?

If not, I'm happy to open a PR and make the updates.

yihui:
Thanks for the report! We'd welcome a PR to use the latest 2.7.x.

> Other than the fact that bookdown is maintained by RStudio, is there a particular reason for pointing to https://mathjax.rstudio.com/latest/?

We started to host MathJax shortly after the original MathJax CDN was shut down a few years ago: https://www.mathjax.org/cdn-shutting-down/ The goal was that https://mathjax.rstudio.com/latest/ always points to the latest version of MathJax, but apparently our IT failed to do so at some point. @cderv Could you file a ticket? Thanks!

cderv:
It is already filled there since some time to move to 3: https://github.com/rstudio/rstudio/issues/8715

I believe updated the hosted version must be synced with IDE update too. AFAIK.

Just recently update to latest 2.7 is on the todo: https://github.com/rstudio/rstudio/issues/11535 as linked by @scarnecchia 

I added a comment so that we are sure to do both.

scarnecchia:
I've implemented the change on a fork, but I'll hold off on creating a PR until I see a response to your question on rstudio/rstudio#11535—it strikes me as more efficient if Rstudio IT can update the hosted version upstream.

cderv:
closed by #1355

cderv:
[...]

PR #1355, opened 2022-06-30
Update gitbook and bs4_book templates to use latest version of MathJax 2.7 from cdnjs

scarnecchia:
This PR implements the changes discussed in #1354 by updating the source for MathJax in inst/templates/bs4_book.html and inst/templates/github.html to use cloudflare's cdnjs latest for 2.7. This closes an XSS vulnerability ([CVE-2018-1999024](https://nvd.nist.gov/vuln/detail/CVE-2018-1999024)) found in versions of MathJax in version's earlier than 2.7.4.

I tested this using a couple of input files I had lying around and I can confirm that the version is now MathJax 2.7.9 and that latex is rendering correctly. I did not see an explicit test for math in the test folder. I'd be happy to add one, but not sure I'll get to it today.


yihui (approved):
Thanks! I noticed that the mode of three files was changed from 644 to 755. Do you know how to revert it?

scarnecchia:
Hi @yihui : I've resolved that issue. File mode should be 644 again (and I've updated my git config to prevent that from happening in the future)

yihui (approved):
Great! We will use https://mathjax.rstudio.com/latest/ again once if actually points to the latest 2.x version. Thank you very much!

PR #1363, opened 2022-07-27
Add example to include a note within citation parentheses in referenc…

acebulsk:
…e to Stack answer https://stackoverflow.com/a/65087765/9609898


yihui (approved):
Thank you!

Issue #764, opened 2019-08-27
Feature request: Support for code folding buttons


bblodfon:
Hi,

I recently searched for an option to have buttons that hide the R code output since it was not supported via a simple option in `bookdown::gitbook`. Using [this stackoverflow answer](https://stackoverflow.com/questions/45360998/code-folding-in-bookdown) I implemented somewhat of a hacked version with nice CSS (from the `united` bootstrap theme) for the buttons and it looks like [this](https://bblodfon.github.io/gitsbe-model-analysis/atopo/cell-lines-2500/a498-model-analysis.html). For the individual R chunks the solution is nice, though for the **hide-show-all** button that is in the right-up corner, the solution is sowewhat (!) hacked (`position: fixed` in the CSS). But it works so it's good enough for me :)

I was considering though that it would be really nice, to have the option integrated globally in gitbook, where the **hide-show-all** button would be in the toolbar (like all the other sharing buttons - github, fb, etc.) and the config would look somewhat like this:

```
bookdown::gitbook:
  config:
    show_codefolding_buttons: true # to show all the buttons
    hide_code: true # initial configuration, no R code is shown if above is true
```

BR, John.

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('bookdown')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('rstudio/bookdown')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

bblodfon:
Based on the previosuly-mentioned [stackoverflow answer](https://stackoverflow.com/questions/45360998/code-folding-in-bookdown), I implemented an R package with a bookdown template that uses a code-folding buttons. Check it here: https://github.com/bblodfon/rtemps.

This is more like a quick-hacked-working solution - I don't know if this should be implemented as part of the main `bookdown` package or not at this point.

cderv:
[...]

PR #1368, opened 2022-08-14
gitbook gains code_folding option (solves #764)

atusy:


atusy:
GitHub Actions with Pandoc devel fails due to GLIBC as reported by https://github.com/rstudio/rmarkdown/issues/2396

yihui (approved):
Great! Thank you!

PR #1385, opened 2022-09-17
Feature: 1380 Sticky Header Update

### LINKED ISSUE #1380: [FR] make the bs4_book sticky menu sticky on mobile
From the bookdown guide about bs4_book: "Thoughtful typography to make the contents as easy as possible to read, regardless of the size of your device. A sticky header gets out of your way when reading, but is easily accessible if you need it.”

The bs4_book menu is not sticky on small screens. (This might be a bug report, given what the documentation says?) 

If I'm reading in the middle of a long page, and I want to go to another part of the book, I have no obvious way on the screen to do so. I have to remember that there was a hamburger menu at the top of the page, and I have to know what it does already. It would be really handy to have a  sticky menu on mobile, especially when you are reading a long page. 

In particular, "sticky" implies to me that the hamburger menu for the TOC floats on the screen, or if it disappears, it shows back up the moment I start scrolling back up toward the top of the page.

ax-jkriese:
Adds functionality for a sticky header for mobile

### DIFF
--- a/inst/resources/bs4_book/bs4_book.css
+++ b/inst/resources/bs4_book/bs4_book.css
@@ -58,7 +58,53 @@ main {margin-top: 1rem;}
     padding-left: 1rem;
   }
   .sidebar-book {
-    margin-top: 1rem;
+    background-color: #fff;
+    padding-top: 1rem;
+    position: relative;
+    transition: .5s ease top;
+    top: -4rem;
+    transform: translateY(4rem);
+    z-index: 1
+  }
+  .sidebar-book .nav-toggle-btn {
+    height: 2.5rem;
+    /*! margin-left: 1rem; */
+  }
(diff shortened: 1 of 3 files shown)


jtbayly:
Is this too risky to merge, or something, @cderv?

I'm trying to decide whether I should wait for it or figure out how to pull in this code when building our books.

cderv:
I want to find time to test it and see the impact it has, especially on current users. As you have tested it, is this something that some users would want to opt-out ? 

My understanding was that it was for mobile only, but the current change seems to also modify the CSS for current style of non mobile view.  I don't want to force changes of style to all existing bs4_book. 

Maybe we need some media query rule to use the CSS based on size of screen ? https://ui.dev/rwd/develop/browser-feature-support/media-queries-for-common-device-breakpoints

PR #1386, opened 2022-09-21
Fix Theorem / Proof environments with special markdown content

### LINKED ISSUE #1371: Theorems env with only a list wrongly number the title as part of the list
See https://community.rstudio.com/t/bookdown-v0-25-exercise-titles-should-not-be-numbered/143961/2

````markdown
---
title: test
output: bookdown::html_document2
---

# Excercises

Do exercises \@ref(exr:exr-1)

::: {#exr-1 .exercise}

1)  Open RStudio.
2)  Write `1 + 9` to console.
:::


::: {#exr-2 .exercise}
\ 

1)  Open RStudio.
2)  Write `1 + 9` to console.
:::

::: {#exr-3 .exercise}

Title 

1)  Open RStudio.
2)  Write `1 + 9` to console.
:::

````

image

See how first one as the title with label be part of the list

### PR DESCRIPTION
**Pull request by cderv**
This fixes #1371 and also make sure we handle other type of content like code chunk

````markdown

::: {#exr-1 .exercise}
1)  Open RStudio.
2)  Write `1 + 9` to console.
:::

::: {.exercise}
```r
library(knitr)
kable(mtcars)
```
Copy and paste the chunk above
:::
````

Adding `\leavevmode` was found in amsthm doc (http://www.ams.org/arc/tex/amscls/amsthdoc.pdf) - it is not the only way of doing it, but it seems to be required to correctly format the result in LaTeX. See explanation in the doc.

@yihui does it seems ok ? Do you think of other cases to test maybe ? I think the solution is general enough to support any usecase.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: bookdown
 Type: Package
 Title: Authoring Books and Technical Documents with R Markdown
-Version: 0.29.1
+Version: 0.29.2
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("JJ", "Allaire", role = "ctb"),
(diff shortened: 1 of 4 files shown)


yihui (approved):
Looks good to me (I didn't know the `\leavevmode` command before). Thank you!

cderv:
> (I didn't know the \leavevmode command before)

Yes me neither. I found it because the result looked awful on PDF !! Hopefully I found this trick 😄

PR #1393, opened 2022-11-18
Fix bug if no edit button

hadley:
Discovered this because I'm trying to turn of editing for R4DS first ed.


hadley:
Please ignore.

PR #1407, opened 2023-02-27
Expand (LaTeX) index section

Blake-Madden:
Mentions the `imakeindex` package and some its options. Also, explain how to add subentries and show LaTeX code for adding a "see also" feature to the index.


yihui (approved):
Thank you very much!

PR #1408, opened 2023-02-27
Fix extra_dependencies handling in gitbook()

ThierryO:
The current implementation ignores `extra_dependencies` passed to `gitbook()`.


cderv (approved):
😮 good finding ! this is really something missing. This means that all of the previous fix we did in https://github.com/rstudio/bookdown/issues/1101 was indeed not working. We should have waited for confirmation before merging. 

I think we need to add a small tests to this PR to confirm. Thanks a lot for this !

ThierryO:
Here is set of [working examples](https://github.com/inbo/inbomd_examples) with bookdown combined with another style . I use this collection of documents to test [our corporate identity](https://github.com/inbo/inbomd). Working on INBOmd is how I found this problem.

cderv:
Awesome. Would that be ok to use this repo in our CI checks so that we can test if a new bookdown version is breaking something ? 
Or maybe you testing is enough ... 🤔

ThierryO:
Sure. That's why we make them public with a open source license.

PR #1421, opened 2023-04-14
Fix for dev testthat

hadley:
Since skip functions now return NULL.


yihui (approved):
Thanks!

hadley:
Just a note that dev testthat will be going to CRAN on May 1 or 2.

yihui:
Thanks for the heads up! We'll try to make a release accordingly.

PR #1423, opened 2023-05-04
Fix default math rendering in epub format

### LINKED ISSUE #1417: Math in epub doesn't work anymore by default
This was reported in https://community.rstudio.com/t/equations-do-not-render-in-epub-version/162380/2

We missed a changed in Pandoc 2.19 

> EPUB writer:

>
> Allow choice of math method for v3 ([#8164](https://github.com/jgm/pandoc/issues/8164)). Previously we always used MathML for math in EPUB3, because the spec includes MathML. But this is not widely supported by readers, so it seems better to allow users to choose their math method as they can with EPUB2 or HTML. NOTE: Existing workflows that produce EPUBv3 documents including math will be affected by this change. You must add --mathml to your command line if you want to continue producing MathML.

Math needs to be added explicitly now. 

- [ ] Also update bookdown-demo

cderv:
Fix breaking change of pandoc 2.19 as math has been made optional while previously it defaulted to `--mathml`

This closes #1417 by only putting back the default behavior prior to Pandoc 2.19 change. 

New feature to optionally opt-out or choose another method will be done in another PR probably.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 - Fix an issue with CSL using hanging indent style in `gitbook()` (thanks, @pablobernabeu, #1422).
 
+- Fix an issue with Pandoc 2.19 not rendering math by default in `epub3` format (#1417).
+
 # CHANGES IN bookdown VERSION 0.33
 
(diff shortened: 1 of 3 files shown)


yihui (approved):
Thanks!

PR #1425, opened 2023-05-04
For pandoc 3.0+, use `--split-level`  when  `chapter_level` is passed in `epub_book`

### LINKED ISSUE #1416: [Pandoc 3] `--epub-chapter-level` is deprecated
> A new option `--split-level` replaces `--epub-chapter-level` and affects both EPUB and chunked HTML output. --epub-chapter-level will still work but is deprecated.

https://github.com/rstudio/bookdown/blob/6d9b147416f94de31fb75113a4b77595633e7492/R/ebook.R#L49

cderv:
For you next release you are preparing @yihui

Closes #1416 

`--epub-chapter-level` is deprecated in favor of `--split-level` and throws a warning. We now use correct argument when pandoc 3 is used.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 - Fix an issue with CSL using hanging indent style in `gitbook()` (thanks, @pablobernabeu, #1422).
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
Thanks!

Issue #504, opened 2017-12-26
Equation labels not properly handled when pandoc 2.0 is used with gitbook output


dataopt:
Here is a minimal example for a minimal book with the following `index.Rmd`

```
---
title: "A Book"
author: "Frida Gomam"
site: bookdown::bookdown_site
documentclass: book
output:
  bookdown::gitbook:
    split_by: section
  bookdown::pdf_book: default
---

# Test labels {-}

Second equation label is not replaced with proper tag when using pandoc 2.0

\begin{equation}
x \in \mathbb{R} (\#eq:real)
\end{equation}
Hello
\begin{equation}
y \in \mathbb{Z} (\#eq:integer)
\end{equation}
```

Screenshot of the generated gitbook:
<img width="794" alt="screen shot 2017-12-26 at 4 55 46 am" src="https://user-images.githubusercontent.com/28999218/34353962-154624a4-e9f9-11e7-8b5b-7cfa60c22a45.png">

yihui:
Could you add empty lines before and after `Hello` so that the equation can be in its own paragraph?

dataopt:
Unfortunately, making the equation a paragraph results in too much space in the latex output.  Below are screenshots of before and after putting empty lines around `Hello`.
<img width="657" alt="screen shot 2017-12-28 at 6 44 47 pm" src="https://user-images.githubusercontent.com/28999218/34425946-71cc7c16-ebff-11e7-866a-4164e4287509.png">
<img width="652" alt="screen shot 2017-12-28 at 6 45 23 pm" src="https://user-images.githubusercontent.com/28999218/34425948-74f8d4f2-ebff-11e7-8041-c9d10bfa5d31.png">

yihui:
The spacing issue is easier to solve. A quick google search: https://tex.stackexchange.com/q/224987/9128

dataopt:
Thanks for the suggestion. Unfortunately, using setlength is more a hack than a solution to the problem since to get the desired spacing, I have to specify a negative value for abovedisplayskip.  As the Rmd file contains perfectly valid syntax, the problem is probably a bug in bookdown.

yihui:
I know exactly what the problem is in this case, and I also know to fix it. I just hope I don't have to touch it, for three reasons: (1) I often have too many other things to do; (2) there is a workaround (i.e. this issue does not really block you); (3) [fixing bugs](https://twitter.com/RichRogersIoT/status/946207872269946880) can introduce new problems.

[...]

PR #1432, opened 2023-05-24
Uses wrap = none for HTML output

### LINKED ISSUE #1430: Text references don't work with multiline paragraphs
[Text references](https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html#text-references) don't seem to work if the text is split in multiple lines. 

Example: Create a new bookdown project and go to `02-cross-refs.Rmd` in the sample book and change the first figure to this:

````
(ref:caption) Here is a nice figure!

```{r nice-fig, fig.cap='(ref:caption)', out.width='80%', fig.asp=.75, fig.align='center', fig.alt='Plot with connected points showing that vapor pressure of mercury increases exponentially as temperature increases.'}
par(mar = c(4, 4, .1, .1))
plot(pressure, type = 'b', pch = 19)
```
````

As expected, text reference works and the caption is rendered where it should be:

eliocamp:
This Closes #1430

### DIFF
--- a/R/bs4_book.R
+++ b/R/bs4_book.R
@@ -57,7 +57,7 @@ bs4_book <- function(theme = bs4_book_theme(),
     self_contained = FALSE,
     theme = NULL,
     template = template,
-    pandoc_args = pandoc_args2(pandoc_args),
+    pandoc_args = pandoc_args2(pandoc_args, wrap = 'none'),
     lib_dir = lib_dir,
     extra_dependencies = c(bs4_book_dependency(theme), extra_dependencies),
     ...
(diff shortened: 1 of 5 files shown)


cderv (changes requested):
it seems this broke some tests - we may need to adapt some processing to this change, especially if what is broken has been added after the `--wrap=preserve` addition. 

Can you have a look to see why the specific test fails ? Otherwise I'll do it soon

eliocamp:
Tests pass on my computer (Pandoc version 2.19.2). Oh boy, this will be fun to debug...

eliocamp:
Ok, partial improvement. I changed it to use wrap=preserve for gitbook, but only because that's the format that's testing #504, apparently. So this doesn't really fix the issue, it just seems to hide it. gitbook will still have #1430 unfixed and #504 might be broken in other formats, if I understand things correctly. 

Pandoc <= 2.16.2 fails a lot, though. No idea why.

PR #1447, opened 2023-11-23
expand repo links

ElenaMetori:


DESCRIPTION
```diff
@@ -78,6 +78,7 @@ Suggests:
     rsconnect (>= 0.4.3),
     servr (>= 0.13),
     shiny,
+    svglite,
```
cderv, line 81: Why svglite is added ? 

Missing suggest ? I don't see where it is used in the PR.

cderv (commented):
Cool ! Thanks. 

Maybe we could some test like with 
https://github.com/rstudio/bookdown/blob/cedaac98f2d4b54a97865e5b050a38fb8938cc7f/tests/testthat/test-bs4_book.R#L70-L83

DESCRIPTION
ElenaMetori: to pass the check, since it's used in `inst/examplesindex.Rmd`, otherwise it doesn't let me make a PR...

cderv: So missing suggest - thanks ! 

As this is for rendering this book only (bookdown demo works ok), I looked into it closer. I believe this is due to change in **knitr** where **svglite** is called (https://github.com/yihui/knitr/commit/4300e4968bc3cb2b02df7c5f0482bd508c4d222e)

And we are indeed asking for **svglite** in our gitbook format for this book 
https://github.com/rstudio/bookdown/blob/cedaac98f2d4b54a97865e5b050a38fb8938cc7f/inst/examples/_output.yml#L1-L2

So this is really a dependency only for the book. We do plan its installation
https://github.com/rstudio/bookdown/blob/cedaac98f2d4b54a97865e5b050a38fb8938cc7f/inst/examples/index.Rmd#L34-L36
but too late compare to **knitr** option resolution. 

So we do need to add it only for our documentation book and not the whole package in there

https://github.com/rstudio/bookdown/blob/cedaac98f2d4b54a97865e5b050a38fb8938cc7f/DESCRIPTION#L93

This is what is used for additional deps when building the book
https://github.com/rstudio/bookdown/blob/cedaac98f2d4b54a97865e5b050a38fb8938cc7f/.github/workflows/Book.yaml#L68-L71

So can you move it ? 

Thanks

PR #1472, opened 2024-07-23
Add config option to change MathJax renderer

bwu62:
Not 100% sure, but based on a small amount of testing, I think this may render more neatly without sacrificing on portability or speed.

I noticed this when typesetting the formula for the sample standard deviation

`$$s=\sqrt{\frac1{n-1}\sum_{i=1}^n(x_i-\bar{x})^2}$$`

The default MathJax renderer of "Common HTML" "fakes" the square root top bar by using a border, which causes annoying misalignment with the radical symbol beside it. It also overall looks a bit more raggedy and not as pretty.

This small config change switches the default renderer to HTML+CSS (thanks @dennissxz for [the tip](https://github.com/jupyter-book/jupyter-book/issues/1174#issuecomment-759857170)).

Seems to be a marginal improvement.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # CHANGES IN bookdown VERSION 0.41
 
+- New `mathjax-config` config option to bs4_book and gitbook to control MathJax config string (thanks, @bwu62, #1472). Currently tested and supported settings:
+  - If empty, defaults to original `TeX-MML-AM_CHTML` which renders all equations in common HTML.
+  - If set to `TeX-AMS-MML_HTMLorMML` renders equations in HTML + CSS (which may look nicer for some equations).
+  - If set to `TeX-MML-AM_SVG` renders equations in SVG.
 
 # CHANGES IN bookdown VERSION 0.40
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # CHANGES IN bookdown VERSION 0.41
 
-- New `mathjax-config` config option to bs4_book and gitbook to control MathJax config string (thanks, @bwu62, #1472). Currently tested and supported settings:
+- New `mathjax-config` option for `bs4_book` and `gitbook` to control MathJax config string (thanks, @bwu62, #1472). The option can be set either in the YAML metadata or as a variable in `pandoc_args`. Currently tested and supported settings:
   - If empty, defaults to original `TeX-MML-AM_CHTML` which renders all equations in common HTML.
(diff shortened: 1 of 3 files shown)


cderv:
Thanks for suggesting this change; 

Changing default is always tricky, because it never comes without impact. It "fixes" the square root syntax you see, but does it break something else maybe ? Hard to test all possibly outcome. 

I don't know why exactly but we use `config=TeX-MML-AM_CHTML` in every other places we have (by doing on quick github search). 

So we should either make this easily customisable maybe, rather than changing default for everyone, which will change the equation looks for all bookdown book building out there. 

@yihui what do you think ?

bwu62:
@cderv Thanks for the reply. I think that makes sense, having this as a nondefault option in the YAML is probably a better idea. Maybe we can also add an option for using SVG as the renderer? I'll work on this a bit later and propose another change if nobody beats me to it.

yihui (approved):
I have no idea about the consequences of changing the default `config` here, so I don't have an opinion. I'm okay with either this PR or a new PR that makes it possible to customize the `config` or the whole mathjax URL. Thanks!

bwu62:
Added new `mathjax-config` config option to bs4_book and gitbook templates to control MathJax config string. I did some preliminary tests and the following usages should be supported:

 - If left empty, defaults to original `TeX-MML-AM_CHTML`  which renders in common HTML
 - If set to `TeX-AMS-MML_HTMLorMML` renders in HTML + CSS
 - If set to `TeX-MML-AM_SVG` renders in SVG

I'm sure other config options are possible but these are just the ones I've uncovered.

Let me know if these look good to you guys or feel free to suggest further changes :)

yihui (approved):
Sounds good to me. Could you add a news item to `NEWS.md`? Then we'll merge this PR. Thanks!

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # CHANGES IN bookdown VERSION 0.41
 
+- New `mathjax-config` config option to bs4_book and gitbook to control MathJax config string (thanks, @bwu62, #1472). Currently tested and supported settings:
```
yihui, line 3:
```suggestion
- New `mathjax-config` option for `bs4_book` and `gitbook` to control MathJax config string (thanks, @bwu62, #1472). The option can be set either in the YAML metadata or as a variable in `pandoc_args`. Currently tested and supported settings:
```

yihui (approved):
Thanks!

Issue #1477, opened 2024-10-16
[FR] Line spacing customisable 


hayden-MB:
Hello! 

Bookdown using the GitBook style via BookML is used in our institution as an accessibility solution. This works well in allowing students to apply their own accommodations or just preferences to the course notes without having to contact the course organiser. However, this menu does not allow the user to change the line spacing. 1.5 or 2 line spacing is a common accommodation for students, and as such it would a useful addition to this menu.

> *What I would want to happen*: that the user can edit the line spacing without outside help
> *Exactly what I would want*: an addition to the 'A' menu on the gitbook style which allows line spacing to be chosen. 

Thank you!!

yihui:
This shouldn't be hard if one knows JavaScript and a little CSS (line spacing is controlled by the CSS property `line-height`), but I don't have time for this task. The place to implement this feature is https://github.com/rstudio/bookdown/blob/main/inst/resources/gitbook/js/plugin-fontsettings.js We will welcome a pull request. Thanks!

hayden-MB:
I will look into doing that. 
Pardon my ignorance, but is there a simple way of checking changes? i.e. viewing an example document locally before pushing.

hayden-MB:
I have written code which I hope would do the job, but without a way of checking I wouldn't want to push. Any advice is appreciated.

PR #1478, opened 2024-10-21
Issue #1477: line spacing customiser

hayden-MB:
This should change the line spacing in the font menu on gitbook style. Happy to change any details of this. Hopefully this is acceptable. Apologies for the tiny back and forth commits, I was testing by pulling code from my repo so had to commit.

PR #1479, opened 2024-10-21
Linespace customise FR

hayden-MB:
Pull request for FR #1477. Tested on bookdown-demo.

Rationalle for options: 1.0, 1.5, 2.0, 2.5 and 3.0 are standard options while 1.7 was the default. 

Happy to discuss any of this.


cderv:
superseeding #1478 - just as note for myself

yihui (commented):
Two questions:

1. What did you change in `inst/resources/gitbook/css/style.css`? I feel it shouldn't be necessary to change this file (I think all relevant CSS can be added to `plugin-fontsettings.css`).
2. Instead of hard-coding a few line height choices, would it be better to only add two buttons (one for increasing, and one for reducing line height, say, by 0.1)? By clicking these buttons, you change the `data-line-height` attribute of the element `.book .book-body .page-inner section`, and in CSS, you set `line-height: attr(data-line-height)`. Then readers will have full freedom of choosing the most comfortable line spacing.

hayden-MB:
1. I removed a line which said 'line-space: 1.7' because it was overwriting the linespace menu. I could maybe do `!important` in `plugin-fontsettings.css` but that feels sligtly clunky. Will do whatever you feel is best though!
2. I can work on this.

hayden-MB:
That should be the toggle option. Let me know what we should do on the CSS issue.

yihui (commented):
1. `plugin-fontsettings.css` appears later than `style.css`, which means for the same CSS selector, `plugin-fontsettings.css` has precedence over `stye.css`, so I don't think you need `!important` (which we should typically avoid using).

2. I didn't mean enumerating `line-height` values in CSS, but using the CSS function `attr()`, which gives you full freedom to use any line height specified in an attribute of the DOM element. I hope that's clearer now.

Thanks!

hayden-MB:
Hi there

1. I would have thought this too, but it only works if I remove this line from the style.css. Not sure how to deal with this.

2. Ah, thank you. I simply mirrored the style from font size which is a list of options, presumably to restrict options to a reasonable range.

I have committed your changes to my fork and tested them and they simply don't work. The spacing does not change with the buttons.

hayden-MB:
I don't think attr() works for line-height.

hayden-MB:
There is an issue that if you click reduce a bunch of times after you have reached minimum, you have to click increase the same number of times to start making a difference. Not sure how to fix this without hardcoding options. This may be why font size was done in this way.

yihui (approved):
You are right. Sorry I forgot that `attr()` currently only works for `content`: https://developer.mozilla.org/en-US/docs/Web/CSS/attr

> There is an issue that if you click reduce a bunch of times after you have reached minimum, you have to click increase the same number of times to start making a difference. Not sure how to fix this without hardcoding options.

When `fontState.spacing <= 10`, we stop decreasing it.

BTW, is it still necessary to remove the default line-height from `style.css`? The `style` attribute usually has the highest precedence and will override most rules defined in external stylesheets.

hayden-MB:
Yes we can now return the css to how it was as there is no changes to even the fontplugin css anymore. Should be that all the changes are in the JS page.

hayden-MB:
Not sure why accepting that change seems to have caused tests to fail? Any ideas?

inst/resources/gitbook/css/plugin-fontsettings.css
```diff
@@ -301,3 +301,4 @@ div.book.color-theme-2 {
   background: #252737;
   font-weight: 600;
 }
+
```
yihui, line 304:
```suggestion
```

inst/resources/gitbook/js/plugin-fontsettings.js
```diff
@@ -60,6 +60,20 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
...
+        fontState.spacing++;
+        saveFontSettings();
+    }
+
+// Decrease line spacing
+    function decreaseSpacing(e) {
+        e.preventDefault();
+        fontState.spacing > 10 && fontState.spacing--;
+        saveFontSettings();
+    }
```
yihui, line 75:
```suggestion
    // Increase or decrease line spacing
    function changeSpacing(e, inc = true) {
        e.preventDefault();
        inc ? fontState.spacing++ : (fontState.spacing > 10 && fontState.spacing--);
        saveFontSettings();
    }
```

inst/resources/gitbook/js/plugin-fontsettings.js
```diff
@@ -74,6 +88,14 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
...
             $book.addClass("color-theme-"+fontState.theme);
         }
+        var lineHeight = Math.max(fontState.spacing / 10, 1);
+        $book.find('section').css('line-height', lineHeight);
+
+        if (lineHeight <= 1) {
+            $(".font-settings .font-reduce").prop('disabled', true);
+        } else {
+            $(".font-settings .font-reduce").prop('disabled', false);
+        }
```
yihui, line 98:
```suggestion
        $(".font-settings .spacing-reduce").prop('disabled', lineHeight <= 1);
```

inst/resources/gitbook/js/plugin-fontsettings.js
```diff
@@ -139,6 +162,17 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
                         text: 'Night',
                         onClick: _.partial(changeColorTheme, 2)
                     }
+                
+                ],
+                [
+                    {
+                    text: 'Spacing -',
+                    onClick: decreaseSpacing
```
yihui, line 170:
```suggestion
                    onClick: e => changeSpacing(e, false)
```

inst/resources/gitbook/js/plugin-fontsettings.js
```diff
@@ -139,6 +162,17 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
...
+                    },
+                    {
+                    text: 'Spacing +',
+                    onClick: increaseSpacing
```
yihui, line 174:
```suggestion
                    onClick: changeSpacing
```

inst/resources/gitbook/js/plugin-fontsettings.js
```diff
@@ -139,6 +162,17 @@ gitbook.require(["gitbook", "lodash", "jQuery"], function(gitbook, _, $) {
                         text: 'Night',
                         onClick: _.partial(changeColorTheme, 2)
                     }
+                
+                ],
+                [
+                    {
```
yihui, line 168:
```suggestion
                    {
                    className: 'spacing-reduce',
```

yihui (commented):
It seems that `style.css` was not restored.

> Not sure why accepting that change seems to have caused tests to fail? Any ideas?

I think you can ignore that. Looks like a change in the dev version of Pandoc irrelevant to this PR. We'll look into it.

Thanks!

yihui (approved):
I've tweaked the rest of minor things and the PR is ready to merge now. Thanks again!

PR #1480, opened 2024-11-13
Change calls to `knitr:::%n%` to `%||%` as `%||%`  is a base function

cgrandin:
The functions `knitr:::%n%` and `%||%` are identical, but `%||%` is in `base` and `%n%` is an internal function of the `knitr` package.

The R documentation states that using triple colon operator inside packages is typically a design mistake (https://stat.ethz.ch/R-manual/R-devel/library/base/html/ns-dblcolon.html) and in this case it can and should be avoided.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -489,6 +489,7 @@ verify_rstudio_version = function() {
 
 str_trim = function(x) gsub('^\\s+|\\s+$', '', x)
 
+if (getRversion() < '4.4.0') `%||%` = function(x, y) if (is.null(x)) y else x
 output_md = function() getOption('bookdown.output.markdown', FALSE)
 
 # a theorem engine for knitr (can also be used for lemmas, definitions, etc)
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -489,8 +489,6 @@ verify_rstudio_version = function() {
 
 str_trim = function(x) gsub('^\\s+|\\s+$', '', x)
```
yihui, line 491:
```suggestion

if (getRversion() < '4.4.0') `%||%` = function(x, y) if (is.null(x)) y else x
```

yihui (changes requested):
Thanks! This operator is available only in R 4.4, so we have to define it for R versions below 4.4 like https://github.com/yihui/litedown/blob/376248c309f005175fa7f81e154f46b72c694a37/R/utils.R#L21

BTW, using `:::` is not a design mistake in this particular case. I'm the maintainer of both **bookdown** and **knitr**, and I'm just stealing my own function from one package to another :)

yihui (approved):

PR #1495, opened 2025-04-03
Web API dependent HTML validation 

### LINKED ISSUE #1494: [Testing] CI ignores changes that break HTML
In the process of working on [this PR](https://github.com/rstudio/bookdown/pull/1490) I discovered that one can introduce a change messes up the HTML quite a bit, and the tests will not spot it.

## MRE
MRE here: https://github.com/katrinabrock/bookdown/pull/1

In the above, I skipped a part of the code responsible for ensuring `div` tags are properly matched when split by sections. So, without this code, html with mismatched tags will be generated even for test rmd here: https://github.com/rstudio/bookdown/tree/main/tests/rmd 

## Expected behavior
IMO - When these `test-rmd.R` runs, ideally, it should check for this kind of issue instead of relying on human reviewers to spot the problem. 

## Actual behavior
I ran the full CI and it came up green. 

## Possible Solution
I have some ideas of how this can be solved (almost done solving locally). If you all agree this is a bug I can post the options, but want to make sure you agree this is something you would want before posting detailed pro and con of different options.

katrinabrock:
Closes #1494

### DIFF
--- a/tests/test-rmd.R
+++ b/tests/test-rmd.R
@@ -15,7 +15,9 @@ if (Sys.getenv('NOT_CRAN') == 'true') local({
     validate_html(list.files("rmd", ".html$", full.names = TRUE))
   )
   if(nrow(html_issues) > 0)
-    stop("HTML issues detected in ", paste(html_issues$file, collapse = ', '))
+    stop(nrow(html_issues),
+         " HTML issues detected:\n  ",
+         paste(html_issues$file, html_issues$messages, sep = ': ', collapse = '\n  '))
 
   # split by section works correctly
   ## id is used for html file name
--- a/tests/test-rmd.R
+++ b/tests/test-rmd.R
@@ -10,14 +10,8 @@ if (Sys.getenv('NOT_CRAN') == 'true') local({
   for (f in list.files('rmd', '[.]Rmd$', full.names = TRUE)) {
     rmarkdown::render(f, envir = globalenv(), quiet = TRUE)
   }
- 
-  html_issues = simplify_html_validation(
-    validate_html(list.files("rmd", ".html$", full.names = TRUE))
-  )
-  if(nrow(html_issues) > 0)
-    stop(nrow(html_issues),
-         " HTML issues detected:\n  ",
-         paste(html_issues$file, html_issues$messages, sep = ': ', collapse = '\n  '))
+
+  validate_html(list.files("rmd", ".html$", full.names = TRUE))
 
   # split by section works correctly
   ## id is used for html file name
--- a/tests/test-rmd.R
+++ b/tests/test-rmd.R
@@ -1,3 +1,4 @@
+source('./testthat/helper-validate_html.R')
 # only run this when NOT_CRAN is true (e.g., on Travis CI)
 if (Sys.getenv('NOT_CRAN') == 'true') local({
   all_files = function() {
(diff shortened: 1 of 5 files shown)


katrinabrock:
This code is actually working. (CI is just complaining because I didn't add `curl` as a dependency, but it's already a transitive dependency with `xfun`.)

I marked it as a draft only because I don't think it should be merged without further discussion. In particular, sourcing a helper from `test-rmd.R` seems like it's against the spirit of `test-rmd.R`'s simplicity.

katrinabrock:
@yihui This is fully working from my perspective. I implemented Option 2 form the options described in #1494.

You can see the tests are passing with no changes to the actual bookdown code. To show they're working, I merged the changes in this PR into the [branch demoing the silent failure](https://github.com/katrinabrock/bookdown/pull/1). You can see [the CI job](https://github.com/katrinabrock/bookdown/actions/runs/14264560650/) is now failing with

```
 Error in eval(quote({ : 
  HTML issues detected in section-1.html, section-1.html, section-2.html, section-2.html, section-3.html, section-3.html, sub2.html, sub2.html, sub2.html, sub2.html, subsection-1.html, subsection-1.html, subsection-1.html, subsection-1.html, subsection-3.html, subsection-3.html, subsection-3.html, subsection-3.html, subsection-footnotes-2.html, subsection-footnotes-2.html, subsection-footnotes-2.html, subsection-footnotes-2.html, test-footnote.html, test-footnote.html
Calls: local -> eval.parent -> eval -> eval -> eval -> eval
Execution halted
```
except for the Mac and Windows jobs on which I skip this test as they don't have docker installed.

The passing of mac and windows jobs demonstrates that with no API and `W3C_MARKUP_VALIDATOR_BASEURL` unset, the tests pass without running the HTML validation so this will be opt-in for bookdown developer's local workflow. It will not impact end users at all because bookdown code itself isn't changing, only tests and CI.

tests/test-rmd.R
```diff
@@ -9,10 +10,21 @@ if (Sys.getenv('NOT_CRAN') == 'true') local({
   for (f in list.files('rmd', '[.]Rmd$', full.names = TRUE)) {
     rmarkdown::render(f, envir = globalenv(), quiet = TRUE)
   }
+ 
+  html_issues = simplify_html_validation(
+    validate_html(list.files("rmd", ".html$", full.names = TRUE))
+  )
+  if(nrow(html_issues) > 0)
+    stop("HTML issues detected in ", paste(html_issues$file, collapse = ', '))
```
yihui, line 18: Can we also show `html_issues$messages` here?

yihui (approved):
This is amazing. Thank you!

tests/test-rmd.R
katrinabrock: Done. I also pushed to the breakage demo branch. You can see the result here: https://github.com/katrinabrock/bookdown/actions/runs/14272976278/job/40010183970?pr=1#step:13:252

katrinabrock:
Glad you like it!

btw - here is what the errors look like when you don't make any exceptions:

```
Error in eval(quote({ : 13 HTML issues detected:
  custom-environments.html: CSS: “border-top”: “solid\9” is not a “color” value.
  custom-environments.html: CSS: “border-bottom”: “solid\9” is not a “color” value.
  custom-environments.html: Attribute “number” not allowed on element “div” at this point.
  section-1.html: Attribute “number” not allowed on element “div” at this point.
  section-2.html: Attribute “number” not allowed on element “div” at this point.
  section-3.html: Attribute “number” not allowed on element “div” at this point.
  sub2.html: Attribute “number” not allowed on element “div” at this point.
  subsection-1.html: Attribute “number” not allowed on element “div” at this point.
  subsection-22.html: Attribute “number” not allowed on element “div” at this point.
  subsection-3.html: Attribute “number” not allowed on element “div” at this point.
  subsection-footnotes-1.html: Attribute “number
```

Probably easy enough to resolve. Might make a good first issue for someone.

PR #1512, opened 2025-11-26
Make publish_book() publish to connect.posit.cloud by default

### LINKED ISSUE #1507: [FR] Support publishing to connect cloud
`publish_book()` should support Connect cloud once supported in rsconnect
- https://github.com/rstudio/rsconnect/pull/1217

yihui:
close #1507

### DIFF
--- a/R/publish.R
+++ b/R/publish.R
@@ -1,49 +1,44 @@
-#' Publish a book to the web
+#' Publish a book to a Posit Connect server
 #'
-#' Publish a book to the web. Note that you should be sure to render all
-#' versions of the book before publishing, unless you have specified
-#' \code{render = TRUE}.
-#'
-#' @inheritParams rsconnect::deploySite
-#'
+#' Publish a book to a Connect Server. By default, you should render the book
+#' locally before publishing.
 #' @param name Name of the book (this will be used in the URL path of the
 #'   published book). Defaults to the \code{book_filename} in
 #'   \code{_bookdown.yml} if not specified.
 #' @param account Account name to publish to. Will default to any previously
 #'   published to account or any single account already associated with
 #'   \code{server}.
-#' @param server Server to publish to (by default beta.rstudioconnect.com but
-#'   any RStudio Connect server can be published to).
-#'
+#' @param server Server to publish to (by default connect.posit.cloud, but any
+#'   Posit Connect server can be published to).
+#' @param ... Other arguments to be passed to [rsconnect::deploySite()].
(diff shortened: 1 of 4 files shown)


yihui:
@cderv Do you mind taking a look at the gh-pages deployment? https://github.com/rstudio/bookdown/blob/86e363126d77ef699f74c0f2a52df9ecc4ad50fc/.github/workflows/Book.yaml#L100-L111

I don't remember when it stopped working, but I'm hoping to pull from gh-pages on connect.posit.cloud so we can still enjoy the automatic publishing from CI. Thanks!

yihui:
I can release the new version of bookdown to CRAN at any time. @cderv Is there anything holding the release at the moment?

R/publish.R
```diff
@@ -1,49 +1,44 @@
...
-
-    # see if they want to configure an account (bail if they don't)
-    message('You do not currently have a bookdown.org publishing account ',
-            'configured on this system.')
-    result = readline('Would you like to configure one now? [Y/n]: ')
-    if (tolower(result) == 'n') return(invisible())
-
-    # configure the account
-    rsconnect::connectUser(server = 'bookdown.org')
+publish_book = function(name = NULL, account = NULL, server = "connect.posit.cloud", ...) {
```
nealrichardson: I'm not an expert on how `rsconnect` works, but IIUC, the old code was relying on the `NULL` behavior to pick an account/credentials in the case where there is only one server configured in your environment. By making `server` non-NULL here, we would break the behavior for anyone who is publishing bookdown to a Posit Connect server like that.

nealrichardson (commented):
A few observations, I'm not sure this is quite right but maybe I'm misreading

R/publish.R
```diff
@@ -1,49 +1,44 @@
...
+  # delete local records of bookdown.org
+  accounts = rsconnect::accounts()
+  x1 = 'bookdown.org' %in% accounts$server
```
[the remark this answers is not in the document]
yihui: The interactive prompting will occur only if bookdown.org was found in the accounts. Since we absolutely want to disable publishing to bookdown.org in the near future, I feel it's more appropriate to do it just here. BTW, for non-interactive R sessions, this will have no effect (`readline()` returns an empty string, making the condition FALSE).

R/publish.R
yihui: Another good point. Amended via e11cbead995894e6f007abf132e9775be81b4d8d.

PR #1513, opened 2025-11-26
Full book is now deployed to gh-pages to be deployed to Connect cloud

cderv:
This changes the workflow, and simplifies the `_render.R`

This also means we don't have a build of the book using pandoc devel anymore to check for problems. 

We would need another branch if we want that. 

Companion to 
- #1512


yihui (approved):
Thanks!

PR #1514, opened 2025-12-01
Make connect.posit.cloud the default if no accounts configured

nealrichardson:
Alternative to #1512 that preserves the existing behavior of `publish_book()` and only suggests connect.posit.cloud instead of bookdown.org if you don't have any accounts configured in rsconnect yet.


yihui (commented):
Thanks! Please see if my response in #1512 makes sense.

PR #1515, opened 2025-12-02
Offer to install RCurl if we detect this is missing but needed

cderv:
due to 
- https://github.com/rstudio/rsconnect/issues/1265 
- https://github.com/rstudio/rsconnect/pull/1266

This is only needed if no **rsconnect** release before a **bookdown** release. 

Without this, the users will face the error shown in https://github.com/rstudio/rsconnect/issues/1265


yihui (approved):
It seems rsconnect 1.7.0 is unlikely to make its way to CRAN tonight, so I'm merging this PR and will prepare the bookdown release later.

PR #1516, opened 2025-12-02
rsconnect 1.6.0 is required for connectCloudUser()

cderv:
@yihui I think this is needed for better user experience

- 1.6.0 is required for the function we use
- rsconnect is only a suggests, so we need to check before using it right ? 

Related to #1515 depending if **rsconnect** 1.6.3 goes on CRAN before bookdown.


yihui (approved):
Looks good. Thanks!

